PR #66, opened 2022-12-20
Feature/exchange rate worker

mgnsharon:


apps/blockchain-monitor/src/monitors.ts
```diff
@@ -0,0 +1,8 @@
+import { MonitorSchedule } from '@notional-finance/monitors';
+import { MonitorJob, exchangeRateMonitor } from '@notional-finance/monitors';
+
+export const monitors: Map<string, MonitorJob[]> = new Map([
+  [MonitorSchedule.EVERY_MINUTE, [exchangeRateMonitor]],
```
mgnsharon, line 5: This is where you add a new job, just add it to the array for the schedule you want.

packages/monitors/src/exchange-rates.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const callMap = new Map<number, AggregateCallList>([
+    [1, { provider: providers['mainnet'], calls: mainnetCalls }],
+    [5, { provider: providers['goerli'], calls: goerliCalls }],
+  ]);
+
+  return callMap;
+}
+
+async function run(opts: JobOptions): Promise<void> {
```
mgnsharon, line 77: Do whatever you'd like in the `run` method, you have access to the environment and context of the worker.
In this case, i'm just logging to datadog.

PR #72, opened 2022-12-27
build: add oracles to monitor

mgnsharon:


packages/monitors/src/exchange-rates-mainnet.ts
```diff
@@ -0,0 +1,79 @@
...
+async function aggregateCalls() {
+  const { provider, calls } = getCalls();
+  const { name } = await provider.getNetwork();
+  const { blockNumber, results } = await aggregate(calls, provider);
+  return { network: name, blockNumber, results };
+}
+
+async function run(opts: JobOptions): Promise<void> {
+  try {
+    provider = new providers.JsonRpcBatchProvider({
```
jeffywu, line 51: Can the mainnet provider be import as well so we can re-use across jobs?

packages/monitors/src/exchange-rates-mainnet.ts
```diff
@@ -0,0 +1,79 @@
...
+          blockNumber,
+          exchangeRate: results[currency],
+        });
+      })
+    );
+    console.log(
+      `Scheduled response: ${JSON.stringify({ network, blockNumber, results })}`
+    );
+  } catch (e) {
+    console.log(e);
```
jeffywu, line 73: Should these log statements go to datadog?

packages/monitors/src/exchange-rates-mainnet.ts
```diff
@@ -0,0 +1,79 @@
...
+  JobOptions,
+  MonitorJob,
+  OracleContract,
+} from './types';
+
+let provider: ethers.providers.JsonRpcBatchProvider;
+
+function getCalls(): AggregateCallList {
+  const contracts: OracleContract[] = [
+    ['eth_usd', '0x5f4ec3df9cbd43714fe2740f5e3616155c5b8419', 8],
```
jeffywu, line 16: Can we also add another attribute to each one of these contracts? It would be either "stable" or volatile" depending on the nature of the exchange rate. All of these are "volatile" with the exception of:

wtbc_btc
usdc_usd
dai_usd

mgnsharon: In the log pipeline, I add the currency_type attribute, you should see usdc_usd & dai_usd currently have the attribute, I can add it to wtbc_btc.

jeffywu: I feel like it is going to be mistake prone to maintain state attributes like that across two systems since it is not obvious that is happening here. I'd prefer to have all the configuration information in the code so that we can see and review it in one place.

packages/monitors/src/exchange-rates-mainnet.ts
mgnsharon: yes, I'll change it to send an error

packages/monitors/src/exchange-rates-mainnet.ts
mgnsharon: The worker now initializes the providers, you can use `getProvider(network)` to use it in the jobs.

packages/monitors/src/exchange-rates-mainnet.ts
mgnsharon: Added `volatilityType` configuration, updated datadog metrics and monitors.

jeffywu (approved):

PR #80, opened 2023-01-17
build: adds grab bag of small improvements

matthew-garrett:
- Adds copy option to wallet address click
- Prevents text from being visible on pull down on mobile
- Adds on enter when attempting to view as account
- Adds N logo to pending transaction screens
- Fixes input jumping to confirmation screens after refresh
- Fixes p descendant bug on side drawers
- Adds highlight on terms of service link in transaction confirmation


mgnsharon (approved):

packages/shared/trade/src/transaction-confirmation/transaction-confirmation.tsx
```diff
@@ -178,10 +183,6 @@ const TermsOfService = styled('p')(
   margin-top: 1rem;
```
jeffywu, line 183: In the future we can probably also switch this TermsOfService component to a Typography one, probably Body or something like that

PR #81, opened 2023-01-18
Feature/kpi monitor

mgnsharon:


apps/blockchain-monitor/src/monitors.ts
```diff
@@ -1,18 +1,17 @@
...
   dev: new Map([
     [MonitorSchedule.EVERY_MINUTE, []],
-    [MonitorSchedule.EVERY_15_MINUTES, [exchangeRateMonitorMainnet]],
+    [
+      MonitorSchedule.EVERY_15_MINUTES,
+      [exchangeRateMonitorMainnet, kpiMonitorMainnet],
+    ],
     [MonitorSchedule.EVERY_HOUR, []],
   ]),
   prod: new Map([
```
jeffywu, line 14: KPI monitor should be scheduled in prod?

packages/durable-objects/src/exchange-rate-store.ts
```diff
@@ -46,7 +46,7 @@ export class ExchangeRateStore {
       if (network && typeof network === 'string') {
         const rates = await this.state.storage.get(network);
         const headers = { 'content-type': 'application/json' };
-        const response = new Response(JSON.stringify(rates), {
+        const response = new Response(JSON.stringify(rates ?? {}), {
```
jeffywu, line 49: Should this return a 404 if the rates object is undefined

packages/logging/src/metric-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+      method: 'POST',
+      body,
+      headers: {
+        'content-type': 'application/json',
+        'dd-api-key': config.apiKey,
+      },
+    };
+
+    const resp = await fetch(config.url ?? DefaultConfig.url, opts);
+    // console.log(`submitMetrics: ${JSON.stringify(resp)}`);
```
jeffywu: Remove the commented code

packages/logging/src/metric-logger.ts
```diff
@@ -0,0 +1,36 @@
...
+    return resp;
+  } catch (e) {
+    console.error(e);
```
jeffywu, line 32: If the fetch request errors out, this will swallow the error and no metric will be submitted. Given how we're logging errors in the workers we won't be alerted to issues.

I think this should either log to datadog or allow the error to be thrown.

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+    monthly: volumes.filter((v) => v.date > lastMonth),
+    total: volumes,
+  };
+
+  recordOverallVolume(volumes);
+  ['daily', 'weekly', 'monthly'].forEach((period) => {
+    recordVolumeByCurrency(volumeByPeriod[period], period);
+  });
+
+  console.log(`KPI: ${JSON.stringify(kpis)}`);
```
jeffywu, line 41: Remove debug logging

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+
+const run = async ({ env }: JobOptions) => {
+  try {
+    const id = env.EXCHANGE_RATE_STORE.idFromName(
+      env.EXCHANGE_RATES_WORKER_NAME
+    );
+    console.log(`ExchangeRate id: ${id} `);
+    const stub = env.EXCHANGE_RATE_STORE.get(id);
+    const req = new Request(
+      `${env.EXCHANGE_RATE_URL}/exchange-rates?network=homestead`
```
jeffywu, line 275: We don't use the "homestead" terminology anywhere in the rest of our code. We should rename homestead => mainnet everywhere.

Also the network here should be in an environment variable

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+    };
+
+    Object.keys(kpis.accounts).forEach((period) => {
+      const { overall, borrowers } = kpis.accounts[period];
+      const overallSeries = {
+        metric: `kpi.accounts.${period}.overall}`,
+        tags: [`period:${period}`, `type:overall`, `chain:mainnet`],
+        points: [
+          {
+            timestamp: Math.round(new Date().getTime() / 1000),
```
jeffywu, line 218: Don't get a new timestamp here, use `nowSeconds` from above so that all the metrics are reported at the same timestamp.

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+  }, {});
+
+  Object.keys(volumeTotal).forEach((key) => {
+    if (key === 'decimals') return;
+    const overallSeries = {
+      metric: `kpi.volume.overall.${key}`,
+      tags: [`type:overall`, `chain:mainnet`, `volumeType:${key}}`],
+      points: [
+        {
+          timestamp: Math.round(new Date().getTime() / 1000),
```
jeffywu, line 60: Use a single timestamp rather than fetching a new one everytime so we get a synchronized recording.

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+function getLocalVolumeKPI(volumes: DailyLendBorrowVolume[]): VolumeKPI {
+  //calculate the total, lend and borrow volumes in usd
+  return volumes.reduce(
+    (acc, v) => {
+      const localValue = BigNumber.from(v.totalVolumeUnderlyingCash);
+      const usdRate = exchangeRatesUSD.get(
+        `${v.currency.underlyingSymbol.toLowerCase()}_usd`
+      );
+      const usdValue = usdRate
+        ? localValue
```
jeffywu, line 143: Throw errors on missing exchange rates

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+    }, {});
+
+    Object.keys(volumeKPI).forEach((key) => {
+      if (key === 'decimals') return;
+      const currencySeries = {
+        metric: `kpi.volume.${period}.${currency}.${key}`,
+        tags: [`type:currency`, `chain:mainnet`, `volumeType:${key}}`],
+        points: [
+          {
+            timestamp: Math.round(new Date().getTime() / 1000),
```
jeffywu, line 114: Same comment re: timestamps

packages/monitors/src/kpi-monitor.ts
```diff
@@ -0,0 +1,330 @@
...
+    await log({
+      message: (e as Error).message,
+      level: 'error',
+      chain: 'mainnet',
+      action: 'kpi.accounts',
+    });
+  }
+}
+
+function setExchangeRatesUSD(rates: ExchangeRate[]) {
```
jeffywu, line 246: This looks ok to me but I think that a unit test or something would make it more clear that the outputs are correct.

jeffywu (changes requested):

apps/blockchain-monitor/src/monitors.ts
mgnsharon: i'll add it once we have it verified

packages/logging/src/metric-logger.ts
mgnsharon: It now throws the error, I will clean up the endpoints along with implementing the api.

packages/monitors/src/kpi-monitor.ts
mgnsharon: I'll handle renaming it in a separate PR.

packages/monitors/src/kpi-monitor.ts
mgnsharon: I'll add unit tests as part of cleanup.

jeffywu (approved):

PR #86, opened 2023-01-26
build: kpi url env var

mgnsharon:


apps/web/src/containers/LandingPageView/components/Sections/TopNotionalStats/index.tsx
```diff
@@ -8,6 +8,7 @@ import iconAnalyticsDashboard from '@notional-finance/assets/icons/icon-analytic
 import iconDiscordSvg from '@notional-finance/assets/icons/icon-discord.svg';
 import { H1, H3, H4 } from '@notional-finance/mui';
 
+const KPIUrl = process.env['NX_KPIS_URL'] || 'https://data.notional.finance';
```
jeffywu, line 11: Do we need to have a hardcoded override like this?

mgnsharon: need no, thought it would be nice to have a fallback incase an environment var wasn't set for some reason.

jeffywu: Since the env var should always be set, maybe we just let the thing fail if it is not set. I'd prefer that versus having it query a different service silently

jeffywu: Ah nevermind, I thought this was inside the worker. This is fine if it is in the web app

jeffywu (approved):

PR #88, opened 2023-01-28
build: create api endpoint for accounts

mgnsharon:


apps/api/src/index.ts
```diff
@@ -18,8 +16,10 @@ export default {
   async fetch(request: Request, env: APIEnv): Promise<Response> {
     const req = request.clone();
     const { url, method, headers } = request;
+
     const path = new URL(url).pathname;
-    if (method !== 'GET' && method !== 'OPTIONS') {
+
+    if (method !== 'GET' && method !== 'OPTIONS' && method !== 'POST') {
```
jeffywu, line 22: Doesn't this allow someone to POST to any endpoint?

packages/durable-objects/src/accounts-do.ts
```diff
@@ -0,0 +1,88 @@
...
+      return new Response((error as Error).message, { status: 400 });
+    }
+  }
+
+  async update(request: Request) {
+    try {
+      const { events }: SentinalRequest = await request.json();
+      if (events && events.length > 0) {
+        const network = events[0].sentinel.network;
+        const accounts = unique(events.map((e) => e.transaction.from));
```
jeffywu, line 38: I don't think this is the right address, the SentinalRequest should have a variable called `account` on it. The address being stored here is the transaction sender (which may not be the actual address we care about).

packages/durable-objects/src/accounts-do.ts
```diff
@@ -0,0 +1,88 @@
...
+        const existingAccounts =
+          (await this.state.storage.get<string[]>(network)) ?? [];
+        if (network && accounts) {
+          const newAccts = (accounts as string[]).filter(
+            (a: string) => !existingAccounts.includes(a)
+          );
+          existingAccounts.push(...newAccts);
+          await this.state.storage.put(network, existingAccounts);
+          return new Response('Accounts Updated', {
```
jeffywu, line 47: This should log to datadog as well?

packages/durable-objects/src/accounts-do.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  async fetch(request: Request): Promise<Response> {
+    try {
+      const req = request.clone();
+      const { method } = request;
+      if (method !== 'POST' && method !== 'GET') {
+        return new Response('Method Not Allowed', { status: 405 });
+      }
+      if (method === 'POST') {
```
jeffywu, line 21: You will probably need to authenticate the request here, right?

apps/api/src/index.ts
mgnsharon: No, just the accounts endpoint, I'm looking at how I can lock it down

packages/durable-objects/src/accounts-do.ts
mgnsharon: That's what the existing auto task was using, it recorded my wallet addresses.  I can double check there isn't another field we should use

packages/durable-objects/src/accounts-do.ts
mgnsharon: I could create some kind of token that we could put in the url as a param possibly

jeffywu: We might also be able to check the `Host` header in the request in addition to the token as a second layer of protection. It seems like Host can be faked if Cloudflare chooses not to verify it, but I think you can test if Cloudflare does indeed verify it by trying to curl the endpoint from your local machine with a faked Host header.

mgnsharon: I looked at the web app firewall briefly, thought maybe we could allow posts only from the sentinel

jeffywu (approved):

PR #93, opened 2023-02-03
build: add support for vault accounts

mgnsharon:


packages/durable-objects/src/types.ts
```diff
@@ -28,3 +28,53 @@ export interface ExchangeRate {
   decimals: number;
   metadata: Record<string, unknown>;
 }
+
+export enum EventSignature {
```
jeffywu, line 32: You should be able to get these from the NotionalABI object from the @notional-finance/contracts library

PR #100, opened 2023-02-20
Feature/vault mobile summary mk 2

matthew-garrett:


packages/features/vault/src/side-drawers/create-vault-position/create-vault-position.tsx
```diff
@@ -70,52 +73,71 @@ export const CreateVaultPosition = () => {
...
-            leverageRatio: Math.floor(leverageRatio * RATE_PRECISION),
-          })
-        }
-        errorMsg={
-          leverageRatioError ||
-          (underMinAccountBorrow
-            ? Object.assign(messages.error.underMinBorrow, {
-                values: { minBorrowSize, borrowAmount },
+    <>
+      <SummaryWrapper>
```
jeffywu, line 77: Do you have to include the mobile vault summary like this in all the different drawers?

PR #101, opened 2023-02-22
build: hooks up new data/context. Fixes bugs

matthew-garrett:


packages/features/vault/src/side-drawers/manage-vault/use-manage-vault.tsx
```diff
@@ -4,31 +4,56 @@ import { FormattedMessage } from 'react-intl';
 
 export function useManageVault() {
   const { pathname: currentPath } = useLocation();
-  const reduceLeverageOptions = [
```
jeffywu, line 7: I feel like the way you have it here is cleaner...one way you can do this is actually pull in eligibleActions in this hook and then do:
```
return {
  reduceLeverageOptions: reduceLeverageOptions.filter((o) => eligibileActions.includes(o.key))
  ...
}
```

packages/features/vault/src/vault-view/vault-summary.tsx
```diff
@@ -14,17 +14,21 @@ import { useContext } from 'react';
...
 export const VaultSummary = () => {
   const theme = useTheme();
   const { state } = useContext(VaultActionContext);
-  const { vaultAddress } = state || {};
-  const { primaryBorrowSymbol, vaultName } = useVault(vaultAddress);
+  const { vaultAddress, vaultConfig } = state || {};
+
+  const vaultName = vaultConfig?.name;
+
+  const { primaryBorrowSymbol } = useVault(vaultAddress);
```
jeffywu, line 30: You should be able to put primaryBorrowSymbol in the vault action context state

packages/features/vault/src/vault-view/vault-view.tsx
```diff
@@ -1,17 +1,35 @@
...
 
 export const VaultView = () => {
-  const { state } = useContext(VaultActionContext);
+  const { updateState } = useContext(VaultActionContext);
+  const { vaultAddress, sideDrawerKey } = useParams<VaultParams>();
   const txnData = useVaultTransaction();
   const showTransactionConfirmation = txnData ? true : false;
-  const { vaultAddress } = state || {};
+
+  useEffect(() => {
```
jeffywu, line 22: Maybe you want to combine these two useEffect calls into one? I'm not sure if we will see race conditions if both action and address change at the same time. Probably better to be safe.

packages/notionable/src/leveraged-vaults/logic/get-init-vault-action.ts
```diff
@@ -86,6 +86,13 @@ export function getInitVaultAction({
     noEligibleMarketsReason,
     minLeverageRatio,
     maxLeverageRatio,
+    primaryBorrowCurrency: vaultConfig.primaryBorrowCurrency,
+    primaryBorrowSymbol: system.getUnderlyingSymbol(vaultConfig.primaryBorrowCurrency),
+    minBorrowSize: vaultConfig.minAccountBorrowSize.toDisplayStringWithSymbol(0),
+    minAccountBorrowSize: vaultConfig.minAccountBorrowSize,
+    strategyName: VaultFactory.resolveStrategyName(vaultConfig.strategy) || '',
+    // TODO: - this getMinDepositRequiredString method is causing a error need to figure out why
```
jeffywu, line 94: You don't need minDepositString so you can leave that off here

packages/notionable/src/leveraged-vaults/vault-action-store.ts
```diff
@@ -44,6 +44,13 @@ interface VaultInputs {
 
 /** These values are calculated on init, when either the account or vault changes */
 interface VaultInitData {
+
```
jeffywu, line 47: Remove this whitespace --- so ugly

packages/features/vault/src/side-drawers/manage-vault/use-manage-vault.tsx
matthew-garrett: I can change this if you like. What we get from using the big object is here is that the manage vault side drawer only needs to have one loop and one block of JSX to display all the options instead of having two that are basically the same.

jeffywu: Ok up to you, I think it's fine either way

PR #103, opened 2023-02-27
feat(vault): adds deposit collateral sidedrawer

matthew-garrett:
- Adds maturity to all vault details tables
- Moves withdraw / exit vault to lower list
- Updates deposit collateral hook to return transaction properties
- Gets deposit collateral side drawer working

### DIFF
--- a/packages/features/vault/src/side-drawers/deposit-collateral/deposit-collateral.tsx
+++ b/packages/features/vault/src/side-drawers/deposit-collateral/deposit-collateral.tsx
@@ -9,25 +9,12 @@ import { messages } from '../messages';
 export const DepositCollateral = () => {
   const {
     updateState,
-    state: {
-      primaryBorrowSymbol,
-      updatedVaultAccount,
-      vaultAddress,
-      buildTransactionCall,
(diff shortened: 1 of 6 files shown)


packages/features/vault/src/side-drawers/deposit-collateral/deposit-collateral.tsx
```diff
@@ -1,48 +1,45 @@
...
+    },
+  } = useContext(VaultActionContext);
+  const transactionData = useDepositCollateral();
+  const canSubmit = buildTransactionCall ? true : false;
+  const currentVaultAddress = vaultAddress || '';
 
   return (
     <VaultSideDrawer
       action={VAULT_ACTIONS.DEPOSIT_COLLATERAL}
       canSubmit={canSubmit}
```
jeffywu, line 26: vaultAction, canSubmit, vaultAddress and updatedVaultAccount can all come from the context directly, maybe we can just put them inside the VaultSideDrawer component directly?

jeffywu (approved):

PR #105, opened 2023-03-03
build: adds withdraw vault and create vault side drawers

matthew-garrett:
- Adds withdraw and exist and create vault side drawers
- Adds generic helpTextValues prop to action sidebar component


packages/features/vault/src/side-drawers/components/vault-side-drawer.tsx
```diff
@@ -48,29 +47,46 @@ export const VaultSideDrawer = ({
...
     ? formatMaturity(vaultAccount?.maturity)
     : '';
 
+  const handleCancel = () => {
+    clearSideDrawer(`/vaults/${vaultAddress}`);
+    if (vaultAction === VAULT_ACTIONS.WITHDRAW_VAULT) {
+      updateState({
+        maxWithdraw: false,
+        withdrawAmount: undefined,
+        withdrawAmountString: '',
```
jeffywu, line 68: Don't set withdrawAmountString here, it should be unused

jeffywu: When the user cancels, I think the vaultAction should be set to undefined every time? Does that happen elsewhere?

packages/features/vault/src/side-drawers/messages.tsx
```diff
@@ -18,7 +18,7 @@ export const messages = {
       description: 'helptext',
     },
     inputLabel: {
-      defaultMessage: '1. How much do you want to withdraw from your vault?',
+      defaultMessage: 'How much do you want to withdraw from your vault?',
```
jeffywu, line 21: Should we just merge the two "messages.tsx" files in this feature shell?

jeffywu (approved):

PR #112, opened 2023-03-20
build: adds analytics dropdown

matthew-garrett:
- Adds analytics dropdown
- Removes About header link (files to be removed later)


packages/shared/web/src/header/analytics-dropdown/use-analytics-dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+          }}
+        />
+      ),
+      description: (
+        <FormattedMessage defaultMessage={`Follow Notional's stats on Dune`} />
+      ),
+    },
+    {
+      title: <FormattedMessage defaultMessage={'Defi Llama'} />,
+      to: 'https://defillama.com/protocol/notional',
```
jeffywu, line 44: @T-Woodward can you decide on what we want to do with this Defi Llama thing?

T-Woodward: It is apparently moot now: https://twitter.com/0xngmi/status/1637955928887533570?s=20

PR #113, opened 2023-03-21
build: refactors resources dropdown

matthew-garrett:
- Refactors resources dropdown based on new mocks
- Removes about and whats up components


packages/shared/web/src/header/connect-community/connect-community.tsx
```diff
@@ -36,7 +38,7 @@ export function ConnectCommunity() {
         <DiscordIcon
           sx={{
             color: theme.palette.background.accentDefault,
-            fontSize: '3rem',
+            fontSize: '5rem',
```
jeffywu, line 41: Can you convert all these font sizes to px and align to the typography if possible?

PR #124, opened 2023-03-31
fix: change to use capacity remaining everywhere

jeffywu:


packages/features/vault/src/components/vault-sub-nav/use-vault-sub-nav.tsx
```diff
@@ -32,6 +32,7 @@ export const useVaultSubNav = () => {
     {
       label: <FormattedMessage defaultMessage={'Full Documentation'} />,
       key: VAULT_SUB_NAV_ACTIONS.FULL_DOCUMENTATION,
+      // TODO: we need to have a registry for this
```
jeffywu, line 35: @matthew-garrett we'll have to figure out a way to figure out this link location as we list more vaults, a problem for tomorrow though

matthew-garrett (approved):

PR #142, opened 2023-04-21
Feature/stats and transparency

matthew-garrett:
- Adds stats and transparency landing page section
- Updates recent audits to include hovers
- Removes connect wallet button from the landing page mobile nav


packages/shared/web/src/header/mobile-navigation/mobile-navigation.tsx
```diff
@@ -80,7 +80,7 @@ export function MobileNavigation({ ...rest }: TabsProps) {
   return window.innerWidth <= theme.breakpoints.values.sm ? (
     <>
       <Box>
-        {!connected && pathname === '/' && (
+        {!connected && pathname !== '/' && (
```
jeffywu, line 83: @matthew-garrett not sure what this change does but just double checking if you actually want it here? It's the opposite of whatever is happening now...

jeffywu (approved):

packages/shared/web/src/header/mobile-navigation/mobile-navigation.tsx
matthew-garrett: This change removes the connect wallet button from the mobile nav on the landing page. It doesn't work on the landing page and it also clutters up the top nav. If we want to keep a button there we could replace it with launch app.

PR #144, opened 2023-04-25
Feature/new hero

matthew-garrett:
- Adds new lottie hero


apps/web/src/containers/LandingPageView/components/hero/images/data-two.json
jeffywu, line 1: @matthew-garrett  Do we want to remove the data blob that is unused?

PR #147, opened 2023-04-28
build: rewrites notional blog component

matthew-garrett:
- Rewrites NotionalBlog component


apps/web/src/containers/LandingPageView/components/notional-blog/notional-blog.tsx
```diff
@@ -0,0 +1,145 @@
+import { useState, useEffect } from 'react';
+import { FormattedMessage } from 'react-intl';
+// import { useInView } from 'react-intersection-observer';
```
jeffywu, line 3: Just remove this if we're not using it

jeffywu (approved):

PR #162, opened 2023-05-11
Feature/page loading

matthew-garrett:
- Removes loadable component and process from all features
- Adds FeatureLoader to MUI and all features. This will allow loading times to be based off of the unique data in each feature instead of a generalized loader that did not work


packages/common/mui/src/lib/feature-loader/feature-loader.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface FeatureLoaderProps {
+  children: ReactElement<any, any>;
+  featureLoaded?: boolean;
+  backgroundColor?: string;
+}
+
+export const FeatureLoader = ({
+  children,
+  featureLoaded = true,
+  backgroundColor,
```
jeffywu, line 15: Should this just inherit from the dark / light setting from the current app theme? What happens if it is unset?

jeffywu (approved):

packages/common/mui/src/lib/feature-loader/feature-loader.tsx
matthew-garrett: This prop is just for the landing page which we force to be dark. Within the rest of the app it is unset and therefor defaults to the light or dark theme setting

jeffywu:
Ok makes sense

PR #164, opened 2023-05-16
Bugfix/beta UI issues

matthew-garrett:
- Fixes vault sub nav bug
- Fixes portfolio screens not opening at the top
- Fixes Vault Details table not updating when coming from the Manage Vault side drawer in Portfolio


packages/notionable/src/leveraged-vaults/vault-action-manager.ts
```diff
@@ -18,20 +19,23 @@ export const loadVaultActionManager = (
   state$: Observable<VaultActionState>
 ): Observable<Partial<VaultActionState>> => {
   const vaultAddress$ = state$.pipe(map((s) => s.vaultAddress));
+  const defaultVaultAction$ = state$.pipe(map((s) => s.vaultAction || s['sideDrawerKey'])) as Observable<VAULT_ACTIONS>;
```
jeffywu, line 22: Can you add a comment here noting that the sideDrawerKey comes from a URL parameter? Otherwise it is a bit mysterious and I don't want to forget.

PR #166, opened 2023-05-20
Feature: React onboard

jeffywu:


packages/notionable/src/global/global-state.ts
```diff
@@ -0,0 +1,21 @@
...
+  process.env['NX_DATA_URL'] || 'https://data-dev.notional.finance';
+
+export interface GlobalState extends Record<string, unknown> {
+  wallet?: {
+    selectedChain?: Network;
+    selectedAddress: string;
+    hasSelectedChainError: boolean;
+  };
+  selectedNetwork?: Network;
+  isNetworkReady: boolean;
```
jeffywu, line 14: We should have a pending state for this as well.

packages/shared/wallet/src/hooks/use-connect.ts
```diff
@@ -0,0 +1,71 @@
...
+  const selectedChain = wallet?.chains[0].id
+    ? getNetworkFromId(BigNumber.from(wallet.chains[0].id).toNumber())
+    : undefined;
+
+  const connectWallet = useCallback(
+    (walletLabel?: string) => {
+      // No change to wallets, nothing to do here.
+      if (!walletLabel || currentLabel === walletLabel) return;
+      connect({ autoSelect: { label: walletLabel, disableModals: true } });
+    },
```
jeffywu, line 32: Issue a pending state when the wallet is connecting

PR #175, opened 2023-05-30
build: adds new wallet connection

matthew-garrett:
- Updates walletSwitcher, walletSelector, ViewAsAccount to use useConnect
- Adds hardcoded wallet modules


packages/shared/wallet/src/view-as-account/view-as-account.tsx
```diff
@@ -2,20 +2,27 @@ import React, { useState } from 'react';
...
   const [error, setError] = useState<boolean>(false);
-  const { setReadOnlyAddress } = useAccount();
   const { clearWalletSideDrawer } = useSideDrawerManager();
+  const { globalState, updateNotional } = useNotionalContext();
 
   const handleClick = () => {
     if (ethers.utils.isAddress(address)) {
-      setReadOnlyAddress(address);
+      updateNotional({
+        ...globalState,
```
jeffywu, line 19: We don't need this `...globalState` I don't think

jeffywu (approved):

PR #176, opened 2023-05-31
build: adds arbitrum to network selectors

matthew-garrett:
- Adds Arbitrum to network selector
- Temporarily sets Arbitrum to default
NOTE* This is a temporary solution and the network selectors do not function fully


packages/shared/wallet/src/settings-side-drawer/network-selector/network-selector.tsx
```diff
@@ -80,7 +90,7 @@ export function NetworkSelector() {
                 alignItems: 'center',
               }}
             >
-              {chain?.id === data.id ? (
+              {chains[0].id === data.id ? (
```
jeffywu, line 93: Why is the zero index used here?

packages/shared/wallet/src/settings-side-drawer/network-selector/use-network-selector.ts
```diff
@@ -11,6 +11,9 @@ export function useNetworkSelector() {
   const { connected, chain } = useOnboard();
   const { getConnectedChain } = useNotional();
   const labels = defineMessages({
+    [CHAIN_NAMES.ARBITRUM]: {
```
jeffywu, line 14: This set of labels looks duplicated with the other set.

packages/shared/config/src/enums.ts
```diff
@@ -110,6 +110,7 @@ export enum NAV_DROPDOWN {
 export enum CHAIN_NAMES {
```
jeffywu, line 110: Remove this and use `Network` instead

packages/shared/wallet/src/network-selector/network-selector.tsx
```diff
@@ -134,7 +42,16 @@ export function NetworkSelector() {
 
   const handleClose = async (chain?: Chain) => {
     if (chain) {
-      switchNetwork(parseInt(chain.id));
+      const currentChain =
```
jeffywu, line 45: There should not be a need to do this parsing if the component uses the `Network` enum rather than some other enum, so you can add a property to the `chains` array which is the `Network` enum value.

packages/shared/wallet/src/onboard-context.tsx
```diff
@@ -16,12 +16,18 @@ import {
...
   {
     id: `0x${NetworkId[Network.ArbitrumOne].toString(16)}`,
     token: 'ETH',
-    label: 'Arbitrum One',
-    rpcUrl: getProviderURLFromNetwork(Network.All),
+    label: 'Arbitrum',
+    rpcUrl: getProviderURLFromNetwork(Network.ArbitrumOne),
+  },
+  {
+    id: `0x${NetworkId[Network.Mainnet].toString(16)}`,
```
jeffywu, line 27: We can just add another key here like `network: Network.ArbitrumOne` and use it to pass to the `updateNotional` call.

packages/shared/wallet/src/settings-side-drawer/network-selector/network-selector.tsx
```diff
@@ -33,23 +32,29 @@ export const NetworkSelectorButton = () => {
...
-      <Box sx={{ marginLeft: theme.spacing(1) }}>
-        {chain?.id && chainEntities[chain.id] && chainEntities[chain.id].label}
-      </Box>
+      <TokenIcon symbol={'arb'} size="medium" />
+      <Box sx={{ marginLeft: theme.spacing(1) }}>{chains[0].label}</Box>
     </Box>
   );
 };
 
 export function NetworkSelector() {
```
jeffywu, line 41: Is there a way that we don't have to duplicate all of this logic in two different files?

packages/shared/wallet/src/settings-side-drawer/network-selector/network-selector.tsx
matthew-garrett: It's hard coded to use Arbitrum for now

packages/shared/wallet/src/network-selector/network-selector.tsx
matthew-garrett: We can't add properties to chains because it breaks onboard. But we can use the label as this property

jeffywu (approved):

PR #195, opened 2023-06-22
build: fixes vault and token allowance bugs

matthew-garrett:
- Fixes token allowance bug to show error message on and and borrow screens when the allowance is too low
- Fixes vault geo blocker bug not working
- Fixes style bug related to CreateVaultPosition

### DIFF
--- a/packages/features/lend/src/lend-sidebar/lend-balance-info.tsx
+++ b/packages/features/lend/src/lend-sidebar/lend-balance-info.tsx
@@ -1,9 +1,15 @@
 import { useEffect } from 'react';
 import { Box, styled, useTheme } from '@mui/material';
-import { MiniButton } from '@notional-finance/mui';
-import { TradeProperty, TradePropertyKeys } from '@notional-finance/trade';
+import { ErrorMessage, MiniButton } from '@notional-finance/mui';
+import {
+  TradeProperty,
+  TradePropertyKeys,
+  tradeErrors,
+} from '@notional-finance/trade';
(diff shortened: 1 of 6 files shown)


packages/features/lend/src/lend-sidebar/lend-balance-info.tsx
```diff
@@ -56,6 +71,18 @@ export const LendBalanceInfo = ({ setInputAmount }: LendBalanceInfoProps) => {
         },
       }}
     >
+      {insufficientAllowance === true && (
```
jeffywu, line 74: @matthew-garrett, FYI i moved the allowance check to here. This is more technically correct, I think,

jeffywu:
@matthew-garrett Can we change the helptext and button text when the vaults are geoblocked?

https://github.com/notional-finance/notional-monorepo/blob/95aac3ca93c06f1001941bff43376ab6796d7f13/packages/features/vault/src/messages.ts

```
    blockedGeoActionHelptext: {
      defaultMessage:
        'Unfortunately, leveraged vaults are not available to U.S. Persons.',
      description: 'error message',
    },
    blockedGeoCTA: {
      defaultMessage: 'Get Clarity for DeFi',
      description: 'button text',
    },
```

The submit button should be this component:
https://github.com/notional-finance/notional-monorepo/blob/dfae0abf8455239a8bfe89e405a52850b100269c/packages/features/vault/src/components/write-congress.tsx

image

jeffywu (approved):

PR #202, opened 2023-06-28
build: fixes txn sidebar bugs

matthew-garrett:
- Refactors txn confirmation flow to use context instead of URL query
- Makes cancel button work on txn confirmation
- Updates Faq to include tokenSymbol
- Fixes other small bugs


packages/features/lend/src/lend-fixed/components/lend-fixed-sidebar.tsx
```diff
@@ -7,38 +7,28 @@ import {
...
+    updateState,
   } = useContext(LendFixedContext);
-  const history = useHistory();
-  const { pathname, search } = useLocation();
   const { currencyInputRef } = useCurrencyInputRef();
 
-  const handleTxnCancel = useCallback(() => {
-    history.push(pathname);
-  }, [history, pathname]);
+  const handleSubmit = () => {
```
jeffywu, line 20: Let's use a useCallback here

PR #210, opened 2023-07-06
Feature/data service

weitianjie2000:


apps/registry/src/index.ts
```diff
@@ -1,10 +1,7 @@
-import {
-  DurableObjectNamespace,
-  Request,
-  Response,
-} from '@cloudflare/workers-types';
+import { DurableObjectNamespace } from '@cloudflare/workers-types';
```
jeffywu, line 1: You might need to still import Request and Response here because they differ slightly from the default types inside the worker environment.

apps/registry/src/index.ts
```diff
@@ -22,6 +19,20 @@ async function runHealthCheck(ns: DurableObjectNamespace, version: string) {
 
 export default {
   async fetch(request: Request, env: RegistryDOEnv): Promise<Response> {
+    const url = new URL(request.url);
+
+    if (url.searchParams.get('blockNumber') !== '') {
```
jeffywu, line 24: Does this require a null check as well in case of no search params or blockNumber is not defined

packages/data-service/src/DataService.ts
```diff
@@ -0,0 +1,197 @@
...
+      )
+      .into(DataService.ORACLE_DATA_TABLE_NAME)
+      .onConflict(['base', 'quote', 'oracle_type', 'network', 'timestamp'])
+      .ignore();
+
+    return blockNumber;
+  }
+
+  private async getRegistryData(blockNumber: number) {
+    const resp = await axios.get(
```
jeffywu, line 147: I'm not sure if axios will work inside a Cloudflare environment. I guess that isn't occuring here but maybe we can just use fetch?

apps/registry/src/index.ts
```diff
@@ -22,6 +19,20 @@ async function runHealthCheck(ns: DurableObjectNamespace, version: string) {
...
+      const network = url.searchParams.get('network') as Network;
+      const blockNumber = parseInt(url.searchParams.get('blockNumber'));
+      const server = new Servers.OracleRegistryServer();
+      const oracles = await server.queryAllOracles(
```
jeffywu, line 28: Instead of calling these two private methods directly can you change this to use a generic `refreshAtBlock` method on `server-registry.js` directly? This will be more generic in the future

```
  // INSIDE base-registry.ts
  protected abstract _refresh(
    network: Network,
    intervalNum: number,
    blockNumber?: number
  ): Promise<CacheSchema<T>>;  

  // INSIDE server-registry.ts
  public async refreshAtBlock(network: Network, blockNumber: number, intervalNum = 0) {
     return  JSON.stringify(await this._refresh(network, intervalNum, blockNumber)));
   }
```

PR #216, opened 2023-07-11
Feature: V3 vault liquidation

weitianjie2000:


.gitignore
```diff
@@ -55,4 +55,4 @@ services/**/dist
 app.yaml
 gcp/data-service/**/dist
 gcp/data-service/**/lib
-gcp/data-service/tsconfig.build.tsbuildinfo
\ No newline at end of file
+gcp/data-service/tsconfig.build.tsbuildinfo
```
jeffywu, line 58: remove `incremental: true` from the tsconfig and then you can delete the tsbuildinfo

packages/contracts/src/abi/BaseStrategyVault.json
```diff
@@ -0,0 +1,513 @@
+[
```
jeffywu, line 1: Is this the same ABI as `IStrategyVault`?

apps/vault-liquidator/src/VaultV3Liquidator.ts
```diff
@@ -0,0 +1,343 @@
+import { ethers, BigNumber, Contract } from 'ethers';
+import { RiskyAccount, IGasOracle } from './types';
+import { AggregateCall, aggregate } from '@notional-finance/sdk/data/Multicall';
```
jeffywu, line 3: Don't use the SDK package Multicall, use the Multicall inside "packages" directly

PR #217, opened 2023-07-11
build: fixes total table row icon issue

matthew-garrett:


packages/common/mui/src/lib/data-table-cells/icon-cell/icon-cell.tsx
```diff
@@ -11,8 +11,18 @@ export interface IconCellProps {
...
   return (
     <Box sx={{ display: 'flex', alignItems: 'center' }}>
-      <TokenIcon symbol={value || 'unknown'} size="medium" />
-      <TableCell style={{ marginLeft: '10px', lineHeight: 'normal' }}>{value}</TableCell>
+      {value !== 'Total' && (
+        <TokenIcon symbol={value || 'unknown'} size="medium" />
+      )}
+      <TableCell
+        style={{
+          marginLeft: value !== 'Total' ? '10px' : '',
```
jeffywu, line 19: Can we export "Total" as a constant from this file so that we don't have to rely on a string match?

PR #239, opened 2023-08-03
Feature/portfolio state zero

matthew-garrett:
- Adds portfolio overview state zero with and without wallet connected
- Fixes feature loader placement
- Fixes stateZeroMessage data table error
- Fixes overview side nav option not being selected with the route is /portfolio


packages/common/styles/src/lib/notional-theme/notional-theme.ts
```diff
@@ -353,7 +353,7 @@ function getTypography(themeVariant: PaletteMode, pageLayout?: NotionalPageLayou
         marginBottom: pxToMargin(12),
       },
       link: {
-        fontSize: pxToRem(12),
+        fontSize: pxToRem(14),
```
jeffywu, line 356: Is this an "approved" change to the typography?

jeffywu (approved):

packages/common/styles/src/lib/notional-theme/notional-theme.ts
matthew-garrett: Approved by me lol. I checked and we don't use that "link" anywhere. In typography it's only used by LinkText. Which is not used anywhere but my new component.

PR #257, opened 2023-08-10
Feature: Adding provider data source and additional config entries

weitianjie2000:


gcp/data-service/src/DataWriter.ts
```diff
@@ -1,6 +1,44 @@
...
+            network: v.networkId,
+            timestamp: context.timestamp,
+            block_number: v.blockNumber,
+            decimals: dataConfig.decimals,
+            value: v.value.toString(),
+          };
+        })
+      )
+      .into(context.tableName)
+      .onConflict(['strategy_id', 'variable', 'network', 'timestamp']);
```
jeffywu, line 32: Are these the right column names for on conflict?

weitianjie2000: fixed

PR #260, opened 2023-08-15
Feature/liquidity updates

matthew-garrett:
- Adds data-table-info-box
- Adds data table title sub text prop
- Updates lend fixed, borrow fixed, leveraged liquidity and variable liquidity how it works images
- Adds totals data info boxes to liquidity page 
- Deletes old how it works images


packages/common/icons/src/assets/icons/currencies/fFRAX.svg
jeffywu, line 1: @matthew-garrett we also need to add the nFRAX and FRAX icons? They should be in the design slack

PR #291, opened 2023-08-30
Feature/add nft contract abi

matthew-garrett:
- Adds NFT contract abi and type
- Adds useNftContract hook
- Adds temp beta nft gif overlay


jeffywu (approved):

apps/web/src/containers/TradingContest/hooks/use-nft-contract.tsx
```diff
@@ -0,0 +1,45 @@
...
+export enum BETA_ACCESS {
+  PENDING = 'pending',
+  CONFIRMED = 'confirmed',
+  REJECTED = 'rejected',
+}
+
+export const useNftContract = () => {
+  const selectedNetwork = useSelectedNetwork();
+  const { selectedAddress } = useConnect();
+  const [betaAccess, setBetaAccess] = useState<BETA_ACCESS>(
```
jeffywu, line 17: This is fine for now, but later we will have to gate the entire UI if the beta access is not found. So we will need to put this into global state somehow.

PR #321, opened 2023-09-15
Feature/portfolio risk updates

matthew-garrett:
- Updates portfolio risk table cell styles
- Adds vault liquidation risk nested table


packages/shared/notionable-hooks/src/use-liquidation-prices.tsx
```diff
@@ -138,7 +138,10 @@ function parseUnderlyingLiquidationPrice(
   return {
     // Used on portfolio screen
     exchangeRate: {
-      symbol: icon,
+      symbol:
+        asset.tokenType === 'VaultShare' && threshold?.toUnderlying()?.symbol
```
jeffywu, line 142: can you refactor this to just
`asset.tokenType === 'VaultShare' && threshold ? ...`

That will save a call to `toUnderlying()` which does take some time

PR #328, opened 2023-09-19
Feature/chart and link updates

matthew-garrett:
- Adds APY charts to Variable Borrow and Lend
- Adds APY and TVL charts to Provide liquidity
- Adds APY chart to Performance chart on leveraged liquidity
- Updates links to V3
NOTE* Please ignore changes on the BarChart component for now

### DIFF
--- a/packages/common/mui/src/lib/action-sidebar/action-sidebar.tsx
+++ b/packages/common/mui/src/lib/action-sidebar/action-sidebar.tsx
@@ -16,7 +16,10 @@ import { NotionalTheme } from '@notional-finance/styles';
 
 export interface ActionSidebarProps {
   heading: MessageDescriptor;
-  helptext: any;
+  helptext:
+    | MessageDescriptor
+    | { defaultMessage: string }
+    | { values?: Record<string, unknown> };
   children: React.ReactNode | React.ReactNode[];
   showDrawer?: boolean;
   canSubmit?: boolean;
--- a/packages/common/mui/src/lib/action-sidebar/action-sidebar.tsx
+++ b/packages/common/mui/src/lib/action-sidebar/action-sidebar.tsx
@@ -16,7 +16,7 @@ import { NotionalTheme } from '@notional-finance/styles';
 
 export interface ActionSidebarProps {
   heading: MessageDescriptor;
-  helptext: MessageDescriptor & { values?: Record<string, unknown> };
+  helptext: any;
   children: React.ReactNode | React.ReactNode[];
   showDrawer?: boolean;
   canSubmit?: boolean;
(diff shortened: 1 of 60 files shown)


packages/common/mui/src/lib/bar-chart/bar-chart.tsx
```diff
@@ -1,7 +1,16 @@
...
   BarChartToolTipDataProps,
 } from './bar-chart-tool-tip/bar-chart-tool-tip';
+// import { ONE_WEEK } from '@notional-finance/util';
+import {
+  getDateString,
+  formatNumberAsPercent,
+  formatNumberToDigits,
+  formatNumber,
+} from '@notional-finance/helpers';
+// import { XAxisTick } from './x-axis-tick/x-axis-tick';
```
jeffywu, line 13: Can ew remove unused imports?

packages/shared/helpers/src/rpc-helpers.ts
```diff
@@ -47,4 +47,43 @@ export function getRpcUrl(networkId: number) {
   }
 }
 
+export function getContractUrls(tokenSymbol: string) {
```
jeffywu, line 50: You should not be using hardcoded URLs. Contract addresses are availble on the TokenDefinition object already.

packages/common/mui/src/lib/bar-chart/bar-chart.tsx
jeffywu: Ok thats fine then

packages/shared/helpers/src/rpc-helpers.ts
jeffywu: For the different pages it will be the following value from the state:
Variable Lend / Liquidity / Leveraged Liquidity: Collateral
Variable Debt: Debt
Fixed Lend / Fixed Borrow / Leveraged Liquidity: Notional Address via the constants file

jeffywu, line 50: Can we remove this code?

packages/common/mui/src/lib/action-sidebar/action-sidebar.tsx
```diff
@@ -16,7 +16,7 @@ import { NotionalTheme } from '@notional-finance/styles';
 
 export interface ActionSidebarProps {
   heading: MessageDescriptor;
-  helptext: MessageDescriptor & { values?: Record<string, unknown> };
+  helptext: any;
```
jeffywu, line 19: Let's not use "any" here, the existing MessageDescriptor type should work right?

matthew-garrett: No it would not work

matthew-garrett: I can fuck with it some more but it was being a real pain and not worth the effort for this particular one in my opinion

jeffywu: What is the cause of the type error? I dont think we should be using the `any` type in any core components like this

PR #373, opened 2023-10-03
Feature/table pending state

matthew-garrett:
- Adds table pending states to portfolio txn history and holdings tables
- Adds nft links to contest pages
- Fixes side drawer bug

### DIFF
--- a/packages/common/mui/src/lib/data-table-cells/multi-value-cell/multi-value-cell.tsx
+++ b/packages/common/mui/src/lib/data-table-cells/multi-value-cell/multi-value-cell.tsx
@@ -13,12 +13,11 @@ export const MultiValueCell = ({ cell }): JSX.Element => {
   const FirstValue = column?.expandableTable ? LargeTableCell : TableCell;
   const SecondValue = column?.expandableTable ? TableCell : SmallTableCell;
 
-  const amountPaidPending =
-    column.showLoadingSpinner && row.original.pendingTokenData;
+  const isPending = column.showLoadingSpinner && row.original.pendingTokenData;
 
   return (
     <Box className="multi-value-cell">
-      {amountPaidPending ? (
+      {isPending ? (
         <ProgressIndicator
           circleSize={24}
           sx={{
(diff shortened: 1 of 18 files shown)


packages/common/mui/src/lib/data-table-cells/multi-value-cell/multi-value-cell.tsx
```diff
@@ -12,38 +13,51 @@ export const MultiValueCell = ({ cell }): JSX.Element => {
   const FirstValue = column?.expandableTable ? LargeTableCell : TableCell;
   const SecondValue = column?.expandableTable ? TableCell : SmallTableCell;
 
+  const amountPaidPending =
```
jeffywu, line 16: Can we rename this to be less specific? Shouldn't this just be like isPending or something? It also applies to other cells as well?

packages/common/mui/src/lib/data-table/data-table.tsx
```diff
@@ -193,6 +203,54 @@ export const DataTable = ({
       {tabBarProps && <DataTableTabBar tabBarProps={tabBarProps} />}
 
       {TabComponentVisible && CustomTabComponent && <CustomTabComponent />}
+      {pendingTokenData && pendingTokenData.pendingTokens.length > 0 && (
+        <Box>
+          <ErrorMessage
+            message={
+              pendingMessage || (
+                <FormattedMessage defaultMessage={'action pending...'} />
```
jeffywu, line 211: Do you need to have a default message here? It looks like you can pass an undefined prop here.

packages/features/portfolio/src/containers/portfolio-txn-history/portfolio-txn-history.tsx
```diff
@@ -54,6 +59,10 @@ export const PortfolioTransactionHistory = () => {
         data={txnHistoryData || []}
         columns={txnHistoryColumns}
         filterBarData={dropdownsData}
+        pendingMessage={
+          <FormattedMessage defaultMessage={'Calculating transaction'} />
```
jeffywu, line 63: Does this include the txn hash link?

packages/common/mui/src/lib/data-table/data-table.tsx
jeffywu: Can we pull this into a separate component? This file is getting pretty big

PR #384, opened 2023-10-10
Feature/overview bar chart

matthew-garrett:
- Adds bar chart


packages/common/mui/src/lib/bar-chart/bar-chart-tool-tip/bar-chart-tool-tip.tsx
```diff
@@ -1,77 +1,68 @@
...
-          </Item>
-        )}
-        {barChartToolTipData?.dataPointTwo && dataPointTwo && (
-          <Item
+  return (
+    <ToolTipBox>
+      <Item>
+        <Box component={'span'}>
+          {payload && payload.length > 0
+            ? intl.formatDate(payload[0]?.payload?.timestamp * 1000, {
```
jeffywu, line 21: Can we use `getDateString` instead?

packages/common/mui/src/lib/bar-chart/bar-chart.tsx
```diff
@@ -20,220 +13,111 @@ import {
...
+      return formatNumberAsAbbr(v, 2);
     }
     if (yAxisTickFormat === 'number' && typeof v === 'number') {
       return formatNumber(v, 2);
     }
     return `${v}`;
   };
 
+  const formatDate = (date) => {
+    return intl.formatDate(date * 1000, {
```
jeffywu, line 61: Use `getDateString`

packages/common/mui/src/lib/bar-chart/bar-chart.tsx
```diff
@@ -20,220 +13,111 @@ import {
...
+              />
+            ))}
+          </RechartsBarChart>
+        </ResponsiveContainer>
+      )}
+    </Box>
   );
 };
 
+// NOTE*
```
jeffywu, line 120: Not sure what this comment refers to?

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -0,0 +1,108 @@
...
+} from '@notional-finance/util';
+import { colors } from '@notional-finance/styles';
+import { useThemeVariant } from '@notional-finance/notionable-hooks';
+import { FormattedMessage } from 'react-intl';
+
+export const useTotalsChart = () => {
+  const baseCurrency = useFiat();
+  const [secondsMultiple, setSecondsMultiple] = useState(1.5);
+
+  window.addEventListener('resize', () => {
```
jeffywu, line 18: Should this go inside a useEffect? It looks like a new listener will get added on every render

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (window.innerWidth <= 1152 && secondsMultiple !== 1.2) {
+      setSecondsMultiple(1.2);
+    }
+    if (window.innerWidth > 1200 && secondsMultiple !== 1.5) {
+      setSecondsMultiple(1.5);
+    }
+  });
+
+  const historyData = useAccountHistoryChart(
+    getNowSeconds() - SECONDS_IN_MONTH * secondsMultiple
```
jeffywu, line 28: Can you remove all the default parameters set in the useAccountHistoryChart hook and just pass them all directly here? That way we don't get any hidden surprises later.

packages/common/mui/src/lib/bar-chart/bar-chart.tsx
matthew-garrett: If you want to have the bars in the bar chart be on top of one another you can pass the same stackId

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
matthew-garrett: If you put this in a useEffect then the window data points will not display as the screen gets smaller and only if a user refreshes on the smaller screen size.

packages/common/mui/src/lib/bar-chart/bar-chart.tsx
matthew-garrett: Just a note to remember that since we had other mocks with stacked bars

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
matthew-garrett: If you're fine with that functionality I can update it

jeffywu: Can't you do this?
https://stackoverflow.com/questions/36862334/get-viewport-window-height-in-reactjs

PR #393, opened 2023-10-17
build: adds leverage liquidity cards

matthew-garrett:
- Adds current APY to portfolio overview chart
- Updates leveraged liquidity cards
- Updates vault cards

### DIFF
--- a/packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
+++ b/packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
@@ -161,8 +161,8 @@ const ContentWrapper = styled(Box, {
   display: flex;
   align-items: baseline;
   justify-content: space-between;
-  width: 474px;
-  min-height: 108px;
+  width: ${theme.spacing(59.25)};
+  min-height: ${theme.spacing(13.5)};
   margin-bottom: ${theme.spacing(2)};
   margin-top: ${theme.spacing(3)};
   margin-right: ${theme.spacing(4)};
@@ -190,7 +190,7 @@ const StyledIcon = styled(Box)(
     left: ${theme.spacing(-1)};
     border-radius: 50%;
     border: 10px solid ${colors.white};
-    padding: 34px;
+    padding: ${theme.spacing(4.25)};
     background: ${theme.palette.common.white};
     box-shadow: ${theme.shape.shadowStandard};
     z-index: 1;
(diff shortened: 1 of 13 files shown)


packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
jeffywu, line 1: This component should be called IncentiveLeveraged instead

packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
```diff
@@ -0,0 +1,203 @@
...
+          }}
+        >
+          {buttonText}
+        </Button>
+      </Card>
+    </Link>
+  );
+};
+
+const ContentWrapper = styled(Box, {
```
jeffywu, line 157: Can we export these from some file and share them across the different cards? They look to be repeated in multiple files.

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -97,6 +104,20 @@ export const useTotalsChart = () => {
...
   }
 
+  if (accountApy) {
+    barConfig.push({
+      dataKey: 'currentApy',
+      title: <FormattedMessage defaultMessage="Current APY" />,
+      toolTipTitle: <FormattedMessage defaultMessage="Debts" />,
+      fill: 'transparent',
+      radius: [8, 8, 0, 0],
+      currencySymbol: FiatSymbols[baseCurrency]
```
jeffywu, line 114: Assume that the currencySymbol does not actually have any effect here?

packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
matthew-garrett: They're slightly different and only used in the incentive leveraged and fixed cards

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
matthew-garrett: Correct this was just to please the type

packages/common/mui/src/lib/card-variants/incentive-active/incentive-active.tsx
matthew-garrett: I imported StyledIcon from the fixed card

PR #421, opened 2023-11-01
build: adds faq how it works and tooltip tracking

matthew-garrett:


packages/common/mui/src/lib/faq/faq.tsx
```diff
@@ -33,6 +34,7 @@ export function Faq({
   componentAnswer,
   slug = '',
   sx,
+  onClick,
```
jeffywu, line 37: Is it possible to make this more foolproof by importing the `trackEvent` function here and just pass in the two `type` and `title` props?

packages/features/lend/src/lend-leveraged/hooks/use-lend-leveraged-faq.tsx
```diff
@@ -36,6 +37,7 @@ export const useLendLeveragedFaq = (tokenSymbol?: string) => {
   ];
   const faqs: FaqProps[] = [
     {
+      questionString: 'What is fTokenSymbol?',
```
jeffywu, line 40: Change this to `What is fCash?`

packages/common/mui/src/lib/info-tooltip/info-tooltip.tsx
```diff
@@ -27,12 +28,14 @@ export function InfoTooltip({
   sx,
   iconColor,
   iconSize,
+  onMouseEnter,
```
jeffywu, line 31: Same here question here as the `Faq` component, can we just import `trackEvent` here instead?

PR #427, opened 2023-11-02
build: fixes table titles and card page widths

matthew-garrett:


packages/shared/web/src/card-container/card-container.tsx
```diff
@@ -169,6 +171,7 @@ const StyledCardList = styled('ul')(
   margin-top: -${theme.spacing(24.5)};
   margin-bottom: ${theme.spacing(22)};
   width: 100%;
+  max-width: ${theme.spacing(187.5)};
```
jeffywu, line 174: If you want to wrap a specific number of flex items without changing the HTML structure, you can achieve this using a combination of CSS and pseudo-elements. Here's how you can do it:

1. HTML Structure:
   Keep your existing HTML structure with all the flex items within a single container.

```html
<div class="flex-container">
  <div class="flex-item">1</div>
  <div class="flex-item">2</div>
  <div class="flex-item">3</div>
  <div class="flex-item">4</div>
  <div class="flex-item">5</div>
  <div class="flex-item">6</div>
  <!-- Add more flex items as needed -->
</div>
```

2. CSS:
   Use CSS to style the container and create pseudo-elements to insert line breaks after a specific number of items to achieve the wrapping effect. You can use the `:nth-child` selector to target and style the pseudo-elements. For example, to wrap every 3 items, you can do the following:

```css
.flex-container {
  display: flex;
  flex-wrap: wrap;
  /* Add other flex properties as needed */
}

.flex-item {
  flex: 1;
  /* Add other styling for the flex items */
}

.flex-item:nth-child(3n)::after {
  content: "";
  flex-basis: 100%;
  order: 0;
}
```

In the CSS above, the `::after` pseudo-element is used to insert a line break after every third flex item. The `flex-basis` property with a value of `100%` forces the pseudo-element to take up the full width, which effectively creates a line break and wraps the items.

This approach allows you to wrap a specific number of flex items without altering the HTML structure. You can adjust the `3n` value to control the number of items before wrapping. For example, you can change it to `4n` to wrap every 4 items, or any other value as needed.

matthew-garrett: Seems plausible. I don't think it's a big deal to add another div imo

matthew-garrett: Nah this doesn't work

jeffywu (approved):

PR #439, opened 2023-11-10
Feat/error message updates

matthew-garrett:
- adds errors on blur of currency input
- updates error styles
- Sets a default value for maturity select
- NOTE* the !importants are to prevent jumping when the errors appear and disappear. We still want the rest of the components in the sidebar to have the same margin bottom

### DIFF
--- a/packages/shared/trade/src/asset-input/asset-input.tsx
+++ b/packages/shared/trade/src/asset-input/asset-input.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useState } from 'react';
-import { Box } from '@mui/material';
+import { Box, useTheme } from '@mui/material';
 import {
   CurrencyInput,
   CurrencyInputHandle,
@@ -64,6 +64,7 @@ export const AssetInput = React.forwardRef<
     },
     ref
   ) => {
+    const theme = useTheme();
     const history = useHistory();
     const { state, updateState } = context;
     const { calculateError } = state;
@@ -142,7 +143,7 @@ export const AssetInput = React.forwardRef<
     if (!availableTokens) return <PageLoading />;
 
     return (
-      <Box>
+      <Box sx={{ marginBottom: `${theme.spacing(4)} !important` }}>
         {label ? label : <InputLabel inputLabel={inputLabel} />}
         <CurrencyInput
           ref={ref}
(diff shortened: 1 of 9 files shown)


packages/shared/trade/src/asset-input/asset-input.tsx
```diff
@@ -142,7 +143,7 @@ export const AssetInput = React.forwardRef<
     if (!availableTokens) return <PageLoading />;
 
     return (
-      <Box>
+      <Box sx={{ marginBottom: `${theme.spacing(4)} !important` }}>
```
jeffywu, line 146: What was this fix for?

packages/shared/trade/src/maturity-select/components/maturities.tsx
```diff
@@ -18,6 +18,11 @@ export function Maturities({
   inputLabel,
 }: MaturitiesProps) {
   const theme = useTheme();
+
+  if (!selectedfCashId) {
```
jeffywu, line 22: I guess we need to discuss the logic around this, not sure if we should select the longest dated maturity by default

packages/shared/trade/src/maturity-select/components/maturities.tsx
```diff
@@ -34,7 +39,7 @@ export function Maturities({
               }
               onSelect={(key) => {
                 const selected =
-                  key && key !== selectedfCashId ? key : undefined;
+                  key && key !== selectedfCashId ? key : selectedfCashId;
```
jeffywu, line 42: If key === selectedfCashId just don't call `onSelect` instead.

packages/shared/trade/src/asset-input/asset-input.tsx
matthew-garrett: NOTE* the !importants are to prevent jumping when the errors appear and disappear. We still want the rest of the components in the sidebar to have the same margin bottom

jeffywu: ok

packages/shared/trade/src/maturity-select/components/maturities.tsx
matthew-garrett: The selected prop does not have access to the key property. It uses selectedfCashId which is why we want to call onSelect

PR #441, opened 2023-11-14
Grab Bag

matthew-garrett:
- Make the Nav Links for Borrow and Earn Yield go to the variable lend / borrow card pages
- Redirect arbitrum.notional.finance/contest to arbitrum.notional.finance
- Disable blocked vpn block button when wallet is not connected
- Under min borrow error on leveraged vaults
- Users that are blocked via VPN or US Geo should not see any "grouped tokens"

### DIFF
--- a/packages/notionable/src/base-trade/sagas/vault-risk.ts
+++ b/packages/notionable/src/base-trade/sagas/vault-risk.ts
@@ -18,6 +18,7 @@ import {
   map,
 } from 'rxjs';
 import { VaultTradeState } from '../base-trade-store';
+import { selectedNetwork } from '../../global';
 import { comparePortfolio, mergeLiquidationPrices } from './account-risk';
 
 export type VaultAccountRiskSummary = ReturnType<typeof vaultRiskSummary>;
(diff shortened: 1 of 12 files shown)


packages/shared/web/src/header/borrow-dropdown/borrow-dropdown.tsx
```diff
@@ -29,6 +30,7 @@ export function BorrowDropDown() {
     <DropdownButton
       popupId="borrow"
       labelKey={messages[NAV_DROPDOWN.BORROW]}
+      onClick={() => history.push('/borrow-variable')}
```
jeffywu, line 33: Is it possible to wrap this up in a Link component rather than use a push? It should show up better in the DOM

packages/notionable/src/base-trade/sagas/vault-risk.ts
```diff
@@ -97,14 +103,40 @@ export function postVaultAccountRisk(
             ?.interestRate
         );
 
+        const vaultCapacity =
```
jeffywu, line 106: Spacing in this code looks really odd

packages/notionable/src/base-trade/sagas/vault-risk.ts
```diff
@@ -97,14 +103,40 @@ export function postVaultAccountRisk(
...
+        network && vaultAddress
+          ? Registry.getConfigurationRegistry().getVaultCapacity(
+              network,
+              vaultAddress
+            )
+          : undefined;
+
+        let underMinAccountBorrow = false;
```
jeffywu, line 114: Should we put this value into the trade state and then read it back out on the component? No sense in doing the calculation twice...

jeffywu: I think our code would be a lot cleaner if we put all the logic for `useVaultCapacity` and `useVaultActionErrors` inside the observable `vault-manager`, maybe we can pull these changes out and put them in a separate PR to handle that.

packages/notionable/src/base-trade/sagas/vault-risk.ts
matthew-garrett: I copy and pasted this from the use-vault-capacity hook

PR #443, opened 2023-11-15
Feat/grab bag updates

matthew-garrett:
• Make the Nav Links for Borrow and Earn Yield go to the variable lend / borrow card pages
• Redirect arbitrum.notional.finance/contest to arbitrum.notional.finance
• Disable blocked vpn block button when wallet is not connected
• Under min borrow error on leveraged vaults
• Users that are blocked via VPN or US Geo should not see any "grouped tokens"


jeffywu (approved):

packages/notionable/src/base-trade/base-trade-store.ts
```diff
@@ -104,6 +104,8 @@ interface TransactionState {
   calculationSuccess: boolean;
   /** True if the calculations are successful and the risk check has completed */
   canSubmit: boolean;
+  /** True if the vault amount is under the minimum borrow size */
+  underMinAccountBorrow?: boolean;
```
jeffywu, line 108: Move this to the VaultState interface

PR #444, opened 2023-11-16
build: moves vault capacity logic to notionable

matthew-garrett:
- moves vault capacity logic to notionable
- updates components to use capacity props from the Vault Context
- fixes markets table link bug


packages/notionable/src/base-trade/base-trade-store.ts
```diff
@@ -173,6 +180,7 @@ export const initialBaseTradeState: BaseTradeState = {
 export const initialVaultTradeState: VaultTradeState = {
   ...initialBaseTradeState,
   customizeLeverage: true,
+  underMinAccountBorrow: false,
```
jeffywu, line 183: Can you make `overCapacityError: false` by default as well?

packages/notionable/src/base-trade/sagas/vault-capacity.ts
```diff
@@ -0,0 +1,107 @@
...
+    selectedNetwork$: ReturnType<typeof selectedNetwork>
+  ) {
+    return combineLatest([account$, state$, selectedNetwork$]).pipe(    
+        distinctUntilChanged(
+            ([, p], [, c]) =>
+              p.tradeType === c.tradeType &&
+              p.calculationSuccess === c.calculationSuccess &&
+              p.collateralBalance?.hashKey === c.collateralBalance?.hashKey &&
+              p.inputErrors === c.inputErrors &&
+              c.debtBalance === c.debtBalance &&
```
jeffywu, line 20: This should compare on the hashKey instead of on the object itself, that will reduce the number of executions we do here.

jeffywu (approved):

PR #448, opened 2023-11-20
build: updates leveraged liquidity page

matthew-garrett:
- Remove nUSDC APY chart tab
- Remove the lever up toggle
- Add tooltip for organic APY
- Remove price exposure table
- Make the lefthand side the same in default and custom
- Put the how it works in the FAQ?
- Keep Organic & NOTE APY on customize mode
- Lower Default leverage to 3.5x
- Show the formula on all pages except when the user is creating a new position with default terms selected


packages/shared/trade/src/trade-action-summary/trade-action-summary.tsx
```diff
@@ -127,6 +127,9 @@ export function TradeActionSummary({
           }
         />
         <TradeActionTitle value={totalAPY} title={apySuffix} valueSuffix="%" />
+        {liquidityYieldData && (
+          <LiquidityYieldInfo liquidityYieldData={liquidityYieldData} />
+        )}
         {isLeveraged && customizeLeverage && (
```
jeffywu, line 133: Does the formula show up on the Manage Leveraged Liquidity actions, like Adjust Leverage, Withdraw, and Roll Maturity?

PR #451, opened 2023-11-22
build: adds portfolio fixes

matthew-garrett:
- Lightning bolt change
- Add divider line above Total row in Portfolio Holdings table
- Total Earnings in Total row should reflect NOTE incentives as well as organic earnings
- Updates width of portfolio side nav notification dot

### DIFF
--- a/packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
+++ b/packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
@@ -47,7 +47,7 @@ export const TotalEarningsTooltip = ({
               sx={{ display: 'flex' }}
             >
               <Box sx={{ marginRight: theme.spacing(1) }}>NOTE</Box>
-              <Box>{noteEarnings}</Box>
+              {noteEarnings}
             </FirstValue>
             <ValueContainer
               sx={{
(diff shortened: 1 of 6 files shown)


packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
```diff
@@ -20,27 +23,65 @@ interface FirstValueProps {
...
+              <FirstValue
+                theme={theme}
+                isNegative={nonNoteEarnings?.includes('-')}
+              >
+                {nonNoteEarnings}
+              </FirstValue>
+            </ValueContainer>
+          </Box>
+        ) : (
+          <>
```
jeffywu, line 69: For consistency can we use a Box here as well

packages/features/portfolio/src/containers/portfolio-holdings/use-detailed-holdings.tsx
```diff
@@ -189,6 +204,8 @@ export function useDetailedHoldings() {
       }
     );
 
+    console.log({ totals });
```
jeffywu, line 207: Remove the log

packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
jeffywu: What is the difference between the first part of the if statement and the second?

matthew-garrett: The first part is only for the total row and shows note usd and combined organic usd values. The second part of the conditional shows four values: note, note usd, token value and the token value usd.

PR #459, opened 2023-11-29
Feat/currency dropdown

matthew-garrett:


packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -43,6 +46,91 @@ export function useWalletBalanceInputCheck(
...
+function useApyValues(tradeType: string | undefined) {
+  // create a apyData object with a type of a Record with key of string and value of string
+  const apyData: Record<string, string> = {};
+  const {
+    yields: { fCashLend, variableLend, liquidity, fCashBorrow, variableBorrow, leveragedLiquidity },
+    getMax,
+    getMin,
+  } = useAllMarkets();
+
+  if(tradeType === 'LendFixed') {
```
jeffywu, line 58: Can you format all the APYs with 3 decimal places?

Also sometimes you use `toPrecision` instead of `formatNumberAsPercent`, we should use the format method to stay consistent.

Also why is there no space after the if in your statements? You should have the `esbenp.prettier-vscode` extension installed. Otherwise I always get a bunch of linter changes when I update any of your files.

jeffywu: Can we also wrap this in a `useMemo`? I feel like it will get called a lot since it is inside the `deposit-input` component

packages/common/mui/src/lib/currency-input/currency-select.tsx
```diff
@@ -181,4 +195,37 @@ export const EmptyCurrencySelectOption = (theme: NotionalTheme) => (
   </StyledItem>
 );
 
+const StyledItem = styled(OptionUnstyled)(
```
jeffywu, line 198: Does not seem like this matches the mock, there aren't any borders between items and the set height does not match what he wants, it should partially show the next currency on the initial pop up

image

packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -43,6 +46,91 @@ export function useWalletBalanceInputCheck(
...
+}
+
+export function useWalletBalances(
+  tokens: TokenDefinition[] | undefined,
+  tradeType: string | undefined,
+) {
+  const { account } = useAccountDefinition();
+  const apyData = useApyValues(tradeType)
+
+  return tokens?.map((token) => {
```
jeffywu, line 110: I guess we should put a `useMemo` here if anything.

jeffywu (approved):

PR #466, opened 2023-12-04
build: fixes bugs

matthew-garrett:
- Removes unneeded error in token approval flow
- Makes transaction confirmation buttons fixed on small screens
- Prevents maturity options disappearing relating to convert-asset/roll-debt

### DIFF
--- a/packages/features/portfolio/src/side-drawers/convert-asset.tsx
+++ b/packages/features/portfolio/src/side-drawers/convert-asset.tsx
@@ -30,7 +30,7 @@ export const ConvertAsset = () => {
   };
 
   return (
-    <>
+    <Box>
       <Box sx={{ display: collateral === undefined ? 'block' : 'none' }}>
         <SelectConvertAsset context={context} />
       </Box>
@@ -56,6 +56,6 @@ export const ConvertAsset = () => {
           />
         </PortfolioSideDrawer>
       )}
-    </>
+    </Box>
   );
 };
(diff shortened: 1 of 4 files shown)


packages/features/portfolio/src/side-drawers/convert-asset.tsx
```diff
@@ -29,28 +29,33 @@ export const ConvertAsset = () => {
...
-        debtOrCollateral="Debt"
-        prefillMax
-        onBalanceChange={onBalanceChange}
-        context={context}
-        inputRef={currencyInputRef}
-        inputLabel={messages[PORTFOLIO_ACTIONS.CONVERT_ASSET]['inputLabel']}
-      />
-    </PortfolioSideDrawer>
+  return (
+    <>
```
jeffywu, line 33: Can you wrap inside a Box instead of a fragment like this?

packages/features/portfolio/src/side-drawers/roll-debt.tsx
```diff
@@ -29,31 +29,36 @@ export const RollDebt = () => {
...
-        debtOrCollateral="Collateral"
-        context={context}
-        prefillMax
-        onBalanceChange={onBalanceChange}
-        inputRef={currencyInputRef}
-        inputLabel={messages[PORTFOLIO_ACTIONS.ROLL_DEBT]['inputLabel']}
-      />
-    </PortfolioSideDrawer>
+  return (
+    <>
```
jeffywu, line 33: Same here, use a Box instead.

jeffywu (approved):

PR #467, opened 2023-12-06
build: updates portfolio table, claim button, transaction page and co…

matthew-garrett:
- Portfolio Holdings Market APY column
- Holdings Total Earnings tooltip
- Claim Incentives button
- Provide Liquidity & Leveraged Liquidity transaction pages
- Confirmation Screen

### DIFF
--- a/packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
+++ b/packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
@@ -8,14 +8,11 @@ import { FormattedMessage } from 'react-intl';
 
 export interface TotalEarningsTooltipProps {
   toolTipData: {
-    nonNoteEarnings?: string;
-    noteEarnings?: string;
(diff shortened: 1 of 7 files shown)


packages/features/portfolio/src/components/total-earnings-tooltip/total-earnings-tooltip.tsx
```diff
@@ -10,10 +10,12 @@ export interface TotalEarningsTooltipProps {
   toolTipData: {
     nonNoteEarnings?: string;
     noteEarnings?: string;
-    underlyingBaseCurrency?: string;
-    underlying?: string;
-    noteBaseCurrency?: string;
-    note?: string;
```
jeffywu, line 16: If there's no special treatment of each these strings, can we make this more readable by just taking two props:

Something like:
```
perAssetEarnings: { underlying: string, baseCurrency: string }[]
totalEarnings: { totalBaseEarnings: string, totalIncentiveEarningsInBase?: string[] }
```

PR #469, opened 2023-12-07
build: updates repay debt max button

matthew-garrett:


packages/shared/trade/src/hooks/use-max-repay.tsx
```diff
@@ -20,9 +19,16 @@ import { TokenBalance } from '@notional-finance/core-entities';
 export function useMaxRepay(context: BaseTradeContext) {
   const {
     updateState,
-    state: { collateral, maxWithdraw },
+    state: { collateral, maxWithdraw, deposit },
   } = context;
   const fCashMarket = useFCashMarket(collateral?.currencyId);
+  const { account } = useAccountDefinition();
```
jeffywu, line 25: Instead of pulling this from the hook again, the `maxBalance` is already returned from useWalletBalanceInputCheck, you can just destructure it there instead.

PR #471, opened 2023-12-08
Feat/arb incentives

jeffywu:


packages/common/mui/src/lib/card-variants/incentive-leveraged/incentive-leveraged.tsx
```diff
@@ -31,21 +42,85 @@ interface ContentWrapperProps {
...
+    incentiveAPY: number;
+  };
+  secondaryIncentiveData?: {
+    symbol: string;
+    incentiveAPY: number;
+  };
+  isIncentive?: boolean;
+}
+
+export const CardValue = ({
```
jeffywu, line 60: @matthew-garrett this is the hover state, is that right? Can we rename the component to something that explains that? Like "IncentiveLeveragedHoverState"? Also this looks like it is exported, is it used elsewhere? If so then we should put it in its own file.

packages/common/mui/src/lib/card-variants/incentive/incentive.tsx
```diff
@@ -69,37 +89,84 @@ export const Incentive = ({
...
-            textAlign="left"
-            marginBottom={theme.spacing(3)}
-            sx={{
-              color: formattedIncentiveRate.includes('-') ? colors.red : '',
-            }}
-          >
-            {formattedIncentiveRate}
-          </CardInput>
-        </>
+            <Box>
```
jeffywu, line 104: @matthew-garrett Is it also possible to put this in its own component and/or maybe merge it with the other hover state? This tsx block is pretty big.

packages/features/liquidity/src/cards.tsx
```diff
@@ -223,6 +223,8 @@ export const LiquidityLeveragedCardView = () => {
   const { depositTokensWithPositions } = useLeveragedNTokenPositions();
   const allMaxAPYs = useMaxYield();
 
+  console.log({ leveragedLiquidity });
```
jeffywu, line 226: Remove log

packages/core-entities/src/Definitions.ts
```diff
@@ -269,7 +269,11 @@ export interface YieldData {
   feeAPY?: number;
   strategyAPY?: number;
   incentives?: {
-    tokenId: string;
```
jeffywu, line 272: @matthew-garrett Can we just rename this `incentives` prop to be `noteIncentives` We're assuming that in the code below anyway.

apps/web/src/containers/Markets/hooks/use-markets-table.tsx
```diff
@@ -168,19 +198,15 @@ export const useMarketsTable = (
           leverage:
             leveraged && leveraged.leverageRatio ? leveraged.leverageRatio : 0,
           totalTVL: tvl?.toFiat(baseCurrency).toFloat() || 0,
-          noteAPY:
-            incentives &&
-            incentives.length > 0 &&
-            incentives[0]?.incentiveAPY > 0
-              ? incentives[0]?.incentiveAPY
-              : 0,
+          incentiveAPY: getIncentiveApy(incentives, secondaryIncentives),
```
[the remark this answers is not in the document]
matthew-garrett: It's used for formatting

packages/common/mui/src/lib/card-variants/incentive/incentive.tsx
matthew-garrett: There is a lot of refactoring that could be done to these cards but as we are about to replace them is that really worth our time?

packages/features/portfolio/src/containers/portfolio-holdings/use-detailed-holdings.tsx
```diff
@@ -49,6 +51,11 @@ export function useDetailedHoldings() {
           t.nonNoteEarnings = t.nonNoteEarnings.add(
             statement.totalProfitAndLoss.toFiat(baseCurrency)
           );
+
+          // TODO: need to update this return value
```
jeffywu, line 55: @matthew-garrett We would need to return something like what we do in `toolTipData` below so that the total would could show all of the total incentives:

```
                    ...totalIncentiveEarnings.map((i) => ({
                      underlying: i.toDisplayStringWithSymbol(),
                      baseCurrency: i
                        .toFiat(baseCurrency)
                        .toDisplayStringWithSymbol(),
                    })),
                  ],
```

PR #488, opened 2023-12-18
Bugfix/notification bug

matthew-garrett:


packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -128,16 +128,22 @@ export function useWalletBalances(
         token,
         content: { 
           balance: maxBalance?.isPositive() ? maxBalance?.toDisplayStringWithSymbol(3, true) : undefined, 
+          usdBalance: maxBalance?.isPositive() ? maxBalance?.toFiat('USD').toDisplayString() : undefined,
```
jeffywu, line 131: You can use `toFloat()` here instead of `toDisplayString()` and then you won't have to `parseFloat` later

packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -128,16 +128,22 @@ export function useWalletBalances(
...
       };
     }).sort((a, b) => {
-      if (a.content.balance && !b.content.balance) {
-          return -1;
+      const balanceA = a.content.usdBalance;
+      const balanceB = b.content.usdBalance;
+      if (balanceA === undefined && balanceB === undefined) {
+          return 0;
+      }
+      if (balanceA === undefined) {
```
jeffywu, line 140: Can you use else if for clarity here? Otherwise good to merge.

PR #492, opened 2023-12-19
build: adds user balance tracking

matthew-garrett:


packages/notionable/src/global/logic.ts
```diff
@@ -45,13 +46,20 @@ export async function onNetworkPending(selectedNetwork: Network) {
...
       selectedNetwork,
       selectedAccount,
-      () => {
+      (a) => {
+        const tokenBalances = a.balances.filter((b) => b.tokenType === 'Underlying').map((data) => {
+          return {
+            [data.token.symbol]: data.toFloat(),
+          }
+        })
+        identify(a.address, selectedNetwork, walletLabel, JSON.stringify(tokenBalances));
```
jeffywu, line 62: Did you delete the other `identify` call? I think it is being called in this file somewhere else.

matthew-garrett: I didn't see another identify call in the file

packages/notionable/src/global/logic.ts
```diff
@@ -45,13 +46,20 @@ export async function onNetworkPending(selectedNetwork: Network) {
...
+  selectedNetwork: Network,
+  walletLabel: string
 ) {
   const isAccountReady = await new Promise<boolean>((resolve) => {
     Registry.getAccountRegistry().onAccountReady(
       selectedNetwork,
       selectedAccount,
-      () => {
+      (a) => {
+        const tokenBalances = a.balances.filter((b) => b.tokenType === 'Underlying').map((data) => {
```
jeffywu, line 57: You should be using a reduce instead of a map here

jeffywu (approved):

PR #494, opened 2023-12-20
Refactor rewards bot

sbuljac:


apps/rewards/src/index.ts
```diff
@@ -51,108 +73,118 @@ const claimRewards = async (env: Env, provider: any) => {
...
+      } while (tradeData.buyAmount == undefined && retryCount++ < 5);
+
+      oracleSlippagePercentOrLimit =
+        BigNumber.from(tradeData.buyAmount).mul(95).div(100).toString();
+      exchangeData = tradeData.data;
+    }
+
+    return [
+      sellToken,
+      token == ARB_WETH ? "0x0000000000000000000000000000000000000000" : token,
```
jeffywu, line 143: It might be cleaner to specify '0x00000..." as ARB_ETH and then whitelist it into the rETH/WETH pool since Balancer handles ETH / WETH conversion automatically. In the future since some Curve pools do not do this for you we may need to ask the contract to sell for WETH explicitly.

apps/rewards/src/index.ts
```diff
@@ -51,108 +73,118 @@ const claimRewards = async (env: Env, provider: any) => {
...
+        );
+        if (tradeData.buyAmount == undefined) {
+          console.log('Request to 0x failed: ', retryCount);
+        }
+        // 0x API sometimes returns  insufficient  liquidity error
+        // so we retry here
+      } while (tradeData.buyAmount == undefined && retryCount++ < 5);
+
+      oracleSlippagePercentOrLimit =
+        BigNumber.from(tradeData.buyAmount).mul(95).div(100).toString();
```
jeffywu, line 137: Can we make this a constant at the top of the file and reduce it to 1%?

apps/rewards/src/index.ts
```diff
@@ -163,55 +195,57 @@ const reinvestToken = async (
...
 
 export default {
-  async fetch(
-    request: Request,
-    env: Env,
-    _: ExecutionContext
-  ): Promise<Response> {
+  async fetch(request: Request, env: Env, _: ExecutionContext): Promise<Response> {
+    const { searchParams } = new URL(request.url)
+    const authKey = searchParams.get('authKey')
```
jeffywu, line 235: Can we pass the key via a request header instead?

jeffywu (approved):

PR #496, opened 2023-12-20
build: fixes overflow

matthew-garrett:
- Fixed overflow bug on fixed cards


packages/common/mui/src/lib/card-variants/currency-fixed/currency-fixed.tsx
```diff
@@ -38,7 +39,7 @@ export const CurrencyFixed = (props: CurrencyFixedProps) => {
   const [hovered, setHovered] = useState(false);
   // NOTE: limits digits to 2 decimal places because of the longer APY suffix, higher
   // apy rates will get cut off.
-  const formattedRate = `${rate.toPrecision(3)}%`;
+  const formattedRate = formatNumberAsPercent(rate, 2);
```
jeffywu, line 42: This will overflow the width when the rates are above 10%, it should be `rate.toFixed(3)`

PR #499, opened 2023-12-22
feat: batch trades for all reward tokens per vault

sbuljac:
Related to https://github.com/notional-finance/staked-note/pull/30


apps/rewards/src/index.ts
```diff
@@ -184,38 +184,51 @@ const getTrades = async (
...
-    env,
-    sellToken,
-    vault.poolTokens,
-    sellAmountsPerToken,
-  );
+  const tradesPerRewardToken = [];
+  for (const sellToken of vault.rewardTokens) {
+    const amount = await ERC20__factory.connect(sellToken, provider).balanceOf(vault.address);
+    console.log("amout is: ", amount);
+    if (amount.lt(100)) {
```
jeffywu, line 197: Can you make DUST amount a constant?

apps/rewards/src/index.ts
```diff
@@ -184,38 +184,51 @@ const getTrades = async (
...
+      sellAmountsPerToken,
+    );
+    tradesPerRewardToken.push(trades);
+  }
+  if (tradesPerRewardToken.length == 0) {
+    console.log(`Found no reward tokens on vault ${vault.address}, skipping.`);
+    return null;
+  }
 
   const data = TreasuryManager__factory.connect(
```
jeffywu, line 218: Can you run a `staticCall` against this method and get the return values? It should return the pool claim amounts that each reinvestment will generate. Then you can use those values and discount them by 1% as inputs here.

PR #505, opened 2024-01-03
build: updates maturity sorting

matthew-garrett:


packages/shared/trade/src/borrow-terms-dropdown/use-borrow-terms.tsx
```diff
@@ -49,7 +49,9 @@ export const useBorrowTerms = (
   const borrowOptions = useMemo(() => {
     // Note this is a any[] because there were conflicts when attempting to combine the debtOptions and spotMaturityData types
     const options = (debtOptions || spotMaturityData) as any[];
-    return options.reverse().map((o, index) => {
+    const variableDebt = options.pop();
+    options.unshift(variableDebt);
```
jeffywu, line 53: Can you add a comment here about what this does? Basically moves the variable debt option to the front of the array

jeffywu (approved):

PR #512, opened 2024-01-09
Feat/health factor changes

matthew-garrett:
- Updates SliderBasic to render the gradient only to one side of the slider
- Updates hideThumb to showThumb since there is only one case for doing that and sets a default value to false
- Updates Liquidation Risk Table to hide the arrow in some instances
- Updates Liquidation Risk Table to show the health factor values based on the new scale
- Adds tooltip cell

### DIFF
--- a/packages/common/mui/src/lib/data-table-cells/arrow-indicator-cell/arrow-indicator-cell.tsx
+++ b/packages/common/mui/src/lib/data-table-cells/arrow-indicator-cell/arrow-indicator-cell.tsx
@@ -3,25 +3,7 @@ import { Box, useTheme } from '@mui/material';
 import { SmallTableCell } from '../../typography/typography';
 import { colors } from '@notional-finance/styles';
 
-export interface ArrowIndicatorCellData {
-  value: string;
-  arrowUp: boolean | null;
-  checkmark: boolean;
-  greenOnArrowUp: boolean;
-  greenOnCheckmark: boolean;
-  isNegative?: boolean;
-}
-
-interface ArrowIndicatorCellProps {
-  cell: {
-    column: any;
-    value: ArrowIndicatorCellData;
-  };
-}
-
-export const ArrowIndicatorCell = ({
-  cell,
-}: ArrowIndicatorCellProps): JSX.Element => {
+export const ArrowIndicatorCell = ({ cell }): JSX.Element => {
   const {
     column: { textAlign, tooRisky },
     value: {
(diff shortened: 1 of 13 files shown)


packages/common/mui/src/lib/data-table-cells/arrow-indicator-cell/arrow-indicator-cell.tsx
```diff
@@ -62,26 +50,36 @@ export const ArrowIndicatorCell = ({
...
-                marginRight: spacing(-0.25),
-                color: tooRisky
-                  ? palette.error.main
-                  : arrowUp
-                  ? colorOnUp
-                  : colorOnDown,
-                width: spacing(1.5),
-                transform: arrowUp ? 'rotate(180deg)' : 'rotate(270deg)',
+                color: isHealthFactor
+                  ? cell?.row?.original?.textColor
```
jeffywu, line 56: Can we remove the `isHealthFactor` boolean and just do `cell?.row?.original?.textColor || valueTextColor` something like that?

packages/shared/notionable-hooks/src/use-summary.tsx
```diff
@@ -757,28 +766,59 @@ export function usePortfolioLiquidationRisk(state: TradeState) {
...
+      content: { defaultMessage: 'Health Factor' },
+      toolTipContent: {
+        defaultMessage:
+          'Your health factor shows your risk. A lower health factor means you have more risk. If your health factor drops below 1, you can be liquidated.',
+      },
+    }),
+    current: onlyCurrent ? updatedValue : _h?.current?.toFixed(2) || 'No Risk',
+    updated: updatedValue,
+    textColor: textColor,
+    hideArrow: !onlyCurrent && priorAccountNoRisk ? true : false,
```
jeffywu, line 808: This doesn't use the `hideArrow` variable above?

packages/shared/notionable-hooks/src/use-summary.tsx
```diff
@@ -757,28 +766,59 @@ export function usePortfolioLiquidationRisk(state: TradeState) {
...
-    current: _h?.current?.toFixed(2) || 'No Risk',
-    updated: _h?.updated?.toFixed(2) || 'No Risk',
+    label: defineMessages({
+      content: { defaultMessage: 'Health Factor' },
+      toolTipContent: {
+        defaultMessage:
+          'Your health factor shows your risk. A lower health factor means you have more risk. If your health factor drops below 1, you can be liquidated.',
+      },
+    }),
+    current: onlyCurrent ? updatedValue : _h?.current?.toFixed(2) || 'No Risk',
```
jeffywu, line 805: I don't think this is right, it should use _h?.current instead of `updatedValue`. Maybe you can make a function that does the formatting for the healthFactor and then pass in either h?.current or h?.updated.

packages/shared/notionable-hooks/src/use-summary.tsx
```diff
@@ -757,28 +766,59 @@ export function usePortfolioLiquidationRisk(state: TradeState) {
...
+
+  const updatedValue =
+    _h?.updated && _h?.updated > 5
+      ? '5+ / 5.0'
+      : !_h?.updated
+      ? 'No Risk'
+      : ` ${_h?.updated?.toFixed(2)} / 5.0`;
+
+  const textColor =
+    _h?.updated && _h?.updated <= HEALTH_FACTOR_RISK_LEVELS.HIGH_RISK
```
jeffywu, line 787: When using `onlyCurrent` the text color needs to refer to h?.current instead

jeffywu (approved):

PR #527, opened 2024-01-19
feat: move rewarder-manager to monorepo

sbuljac:


apps/reward-manager/src/vault_manager.ts
```diff
@@ -0,0 +1,248 @@
+import { Provider } from "@ethersproject/abstract-provider";
```
jeffywu, line 1: We can remove this file since we will use the other reward bot instead.

apps/reward-manager/src/notifier.ts
```diff
@@ -0,0 +1,48 @@
+import DDClient, { DDEvent } from "./dd_client";
```
jeffywu, line 1: Can we remove this in favor of the existing data dog logging package in the monorepo?

apps/reward-manager/src/gas_oracle.ts
```diff
@@ -0,0 +1,63 @@
+import { BigNumber, providers } from "ethers";
```
jeffywu, line 1: I don't think this is really used anymore, this uses the legacy gas price mechanism I believe

apps/reward-manager/src/dd_client.ts
```diff
@@ -0,0 +1,74 @@
+const DD_BASE_URL = "https://api.datadoghq.com";
```
jeffywu, line 1: Same as the notifier.ts file, lets remove this in favor of the existing logging package

apps/reward-manager/src/util.ts
```diff
@@ -0,0 +1,66 @@
+import { BigNumber } from "ethers";
```
jeffywu, line 1: If these are shared methods between the reward bot and this sNOTE reinvestment bot then maybe we can refactor them into the `@notional-finance/utils` package or something.

apps/reward-manager/project.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "reward-manager",
```
jeffywu, line 2: Can we rename this package to `note-reward-manager` or something for clarity? There should be an command like `yarn nx rename` that will make all the changes for you.

apps/reward-manager/src/config/addresses.json
```diff
@@ -0,0 +1,27 @@
+{
+    "goerli": {
+        "treasury": {
+            "address": "0x8638f94155c333fd7087c012Dc51B0528bb06035"
+        },
+        "wstETHVault": {
```
jeffywu, line 6: Any references to the wstETHVault can be removed.

apps/reward-manager/src/config/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+    "mainnet": {
+        "poolIds": {
+            "0xba100000625a3754423978a60c9317c58a424e3d": "0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014",
```
jeffywu, line 4: If these configs are not used in the sNOTE reinvestment then we can delete the file.

apps/reward-manager/src/gas_oracle.ts
sbuljac: will remove it completely since we can set max gas price in defender although currently no cap is set.

apps/reward-manager/src/util.ts
sbuljac: Do I also need to increase semver of util package and publish it to npm now that I added two new functions?

jeffywu (approved):

PR #553, opened 2024-01-27
fix: optimise slippage calculation, fix typos and add new vaults

sbuljac:


packages/util/src/zero-x-exchange.ts
```diff
@@ -33,13 +40,22 @@ export async function get0xData(
     buyToken: buyToken,
     sellAmount: sellAmount.toString(),
     slippagePercentage: String(slippagePercentage / 100)
-  })
+  }).toString();
+
+  console.log(searchParams);
```
jeffywu, line 45: Did you want to keep this log statement?

jeffywu (approved):

PR #561, opened 2024-01-30
Feat/liquidity pool table

matthew-garrett:
- Adds liquidity pool table
- Refactors data table to use data table scroll component when there is a maxHeight prop passed.

*NOTE: This was the best solution to resolve a bug where scrollable tables column headers would not align correctly. It required using the (small) react table sticky library and useBlockLayout hook from react table. This difference was significant enough that it required its own component instead of trying to have the data table component manage everything.


packages/notionable/src/global/global-state.ts
```diff
@@ -56,6 +56,21 @@ const CACHE_HOSTNAME =
 
 export type NetworkLoadingState = 'Pending' | 'Loaded' | undefined;
 
+export type HistoricalTrading = Record<
```
jeffywu, line 59: Is it possible to import this from the analytics server? Maybe you can move all the type definitions at the top of the `analytics-server` file into the `Definitions` file in the core-entities repo and then import from there.

packages/features/liquidity/src/liquidity-variable/hooks/use-liquidity-pools-table.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const {
+    state: { deposit, selectedNetwork },
+  } = useContext(LiquidityContext);
+  const {
+    globalState: { historicalTrading },
+  } = useNotionalContext();
+  let poolTableData: Record<string, any>[] = [];
+
+  if (deposit && deposit?.currencyId && selectedNetwork && historicalTrading) {
+    poolTableData = historicalTrading[selectedNetwork][deposit?.currencyId]
```
jeffywu, line 34: Can you just do all this data processing in the `on-data-update` method? That way we will only re-calculate all this data periodically rather than on every render.

packages/notionable/src/global/sagas/on-data-update.ts
```diff
@@ -107,3 +108,26 @@ function onActiveAccounts$(global$: Observable<GlobalState>) {
...
+        isAppReady(p.networkState) === isAppReady(c.networkState) &&
+        c.baseCurrency === p.baseCurrency
+    ),
+    filter((g) => isAppReady(g.networkState)),
+    switchMap(() => {
+      return timer(500, 60_000).pipe(
+        map(() => ({
+          historicalTrading: SupportedNetworks.reduce((acc, n) => {
+            if (Registry.getAnalyticsRegistry().isNetworkRegistered(n)) {
+              acc[n] = Registry.getAnalyticsRegistry().getHistoricalTrading(n);
```
jeffywu, line 125: I would process the liquidity pool table data for all currency ids at this level and then you can just read them out of global state.

PR #571, opened 2024-01-31
build: adds capacity remaining values

matthew-garrett:
- Adds capacity remaining to total box row of Lend Fixed, Lend Variable, Provide Liquidity and Leveraged Liquidity

### DIFF
--- a/packages/core-entities/src/client/configuration-client.ts
+++ b/packages/core-entities/src/client/configuration-client.ts
@@ -480,7 +480,7 @@ export class ConfigurationClient extends ClientRegistry<AllConfigurationQuery> {
     if (!pCash.totalSupply) throw Error('pCash total supply not found');
     const currentUnderlyingSupply = pCash.totalSupply.toUnderlying();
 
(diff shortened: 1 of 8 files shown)


packages/core-entities/src/client/configuration-client.ts
```diff
@@ -480,7 +480,7 @@ export class ConfigurationClient extends ClientRegistry<AllConfigurationQuery> {
     if (!pCash.totalSupply) throw Error('pCash total supply not found');
     const currentUnderlyingSupply = pCash.totalSupply.toUnderlying();
 
-    return { maxUnderlyingSupply, currentUnderlyingSupply };
+    return { maxUnderlyingSupply, currentUnderlyingSupply, capacityRemaining: maxUnderlyingSupply.sub(currentUnderlyingSupply) };
```
jeffywu, line 483: Does your eslint put this on a separate line?

matthew-garrett: eslint doesn't auto format in this file. It works in some directories but not others idk what's wrong with it.

PR #588, opened 2024-02-07
build: fixes borrow dropdown bugs and others

matthew-garrett:
Fixed issues where borrow terms dropdown was rendering empty after refresh on vaults and when switching from a managed liquidity position to a new token and selecting custom. These fixes resolved the following tickets: 
- Text regarding maturities for LV
- Blank borrow details option after withdrawing from lev vault

Other resolved tickets: 
- High utilization warning does not show when withdrawing leveraged liquidity
- APYs show up as zero on the earn yield popup
- Update feature loader for card pages to include when yields are ready

### DIFF
--- a/packages/features/liquidity/src/liquidity-leveraged/side-drawers/create-increase-position.tsx
+++ b/packages/features/liquidity/src/liquidity-leveraged/side-drawers/create-increase-position.tsx
@@ -22,25 +22,11 @@ export const CreateOrIncreasePosition = () => {
       debt,
       selectedNetwork,
       deposit,
-      isReady,
-      availableDebtTokens,
     },
-    updateState,
   } = context;
   const { currencyInputRef } = useCurrencyInputRef();
   const { currentPosition, depositTokensWithPositions } =
     useLeveragedNTokenPositions(selectedNetwork, selectedDepositToken);
-
-  if (
-    !debt &&
-    availableDebtTokens &&
-    !currentPosition &&
-    customizeLeverage &&
-    isReady
-  ) {
-    const debt = availableDebtTokens?.find((t) => t.tokenType === 'PrimeDebt');
-    updateState({ debt });
-  }
 
   return (
     <TransactionSidebar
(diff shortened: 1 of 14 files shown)


packages/shared/web/src/header/invest-and-earn/use-invest-earn-links.tsx
```diff
@@ -42,10 +42,22 @@ export const useInvestEarnLinks = () => {
         <FormattedMessage
           defaultMessage="Earn up to {rate} APY"
           values={{
-            rate: formatNumberAsPercent(fCashLend?.totalAPY || 0),
+            rate: fCashLend?.totalAPY ? (
```
jeffywu, line 45: Can you make the comparison for all of these `fCashLend?.totalAPY === undefined`? If the value is actually zero then we will end up with an endless spinner.

Also these need to be added to the Borrow dropdown as well. I think we can make a little helper function for it to reduce the code.

packages/features/liquidity/src/liquidity-leveraged/side-drawers/create-increase-position.tsx
```diff
@@ -22,12 +22,26 @@ export const CreateOrIncreasePosition = () => {
...
+      isReady,
+      availableDebtTokens,
     },
+    updateState,
   } = context;
   const { currencyInputRef } = useCurrencyInputRef();
   const { currentPosition, depositTokensWithPositions } =
     useLeveragedNTokenPositions(selectedNetwork, selectedDepositToken);
 
+  if (
```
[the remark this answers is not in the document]
matthew-garrett: This does not work in a useEffect. Without using a useEffect the condition only runs twice on initial page load or on refresh. It does not run when the a user interacts with the page.

packages/notionable/src/base-trade/sagas/tokens.ts
```diff
@@ -206,7 +206,9 @@ export function availableTokens(
         availableDebtTokens.map((t) => t.id).join(':') !==
           s.availableDebtTokens?.map((t) => t.id).join(':') ||
         availableDepositTokens.map((t) => t.id).join(':') !==
-          s.availableDepositTokens?.map((t) => t.id).join(':');
+          s.availableDepositTokens?.map((t) => t.id).join(':') || 
+        s.debt === undefined || s.collateral === undefined
```
jeffywu, line 210: Can we move the hasChanged variable lower after `getSelectedToken` is called and change this comparison to `s.debt?.id !== debt.id || s.collateral?.id !== collateral.id`

PR #622, opened 2024-02-22
Feat/contest leaderboard updates

matthew-garrett:
- Fixes current user table issues
- Adds error when current user is under $100 deposit
- Filters out users from the main contest tables who's deposit amount is less than $100
- Adds partners button bar filtering


apps/web/src/containers/TradingContest/hooks/use-leaderboard-data.ts
```diff
@@ -47,59 +103,51 @@ export function useLeaderboardData() {
...
-            totalAPY: formatNumberAsPercent((a.irr || 0) * 100),
-            totalEarnings: `$${formatNumber(a.earnings)}`,
-            totalDeposits:
-              a.netDeposits < 0 ? `$0.0000` : `$${formatNumber(a.netDeposits)}`,
-            netWorth: `$${formatNumber(a.totalNetWorth)}`,
-          }))
-      );
+      setFatCatData(filteredFatCatData);
+
+      if (userData && userData.netDeposits < 100) {
```
jeffywu, line 124: Instead of `netDeposits` this should be `totalNetWorth`. That applies for all of the places where you are filtering for `netDeposits`, sorry if that was unclear.

PR #624, opened 2024-02-23
Feat/dashboard view

matthew-garrett:
- Adds dashboard view to all former card pages
- Removes old card pages


packages/notionable/src/global/account/holdings.ts
```diff
@@ -200,6 +200,13 @@ export function calculateGroupedHoldings(
                 ? debtHoldings.marketYield.totalAPY
                 : // Need to check for undefined here if the debtHoldings is undefined
                   debtHoldings?.statement?.impliedFixedRate,
+            totalLeveragedApy: leveragedYield(
+              assetHoldings.marketYield?.totalAPY,
+              (debtHoldings?.marketYield?.token.tokenType === 'PrimeDebt'
```
jeffywu, line 205: Can you make this a variable instead of re-computing it here?

packages/shared/web/src/dashboard-view/hooks/use-borrow-fixed-dashboard.tsx
```diff
@@ -0,0 +1,56 @@
...
+        }),
+        apy: y.totalAPY,
+        routeCallback: () =>
+          history.push(
+            `/${PRODUCTS.BORROW_FIXED}/${network}/${y.underlying.symbol}`
+          ),
+      };
+    })
+    .sort((a, b) => a.apy - b.apy)
+    .filter((data) => {
```
jeffywu, line 34: I think we will probably want to sort by TVL initially and that will break this filtering, so you can use a `reduce` instead here. Something like this:

```
const allData = Object.values(fCashBorrow.reduce((acc, y) => {
   if (acc[y.underlying.symbol] === undefined || acc[y.underlying.symbol].apy > y.totalAPY) {
       acc[y.underlying.symbol] = { obj }
   }
}, {})).sort((a, b) => b.tvl - a.tvl)
```

packages/shared/web/src/dashboard-view/hooks/use-borrow-variable-dashboard.tsx
```diff
@@ -0,0 +1,42 @@
...
+        subTitle: `TVL: ${y.tvl ? formatNumberAsAbbr(y.tvl.toFloat(), 0) : 0}`,
+        hasPosition: false,
+        apy: y.totalAPY,
+        routeCallback: () =>
+          history.push(
+            `/${PRODUCTS.BORROW_VARIABLE}/${network}/${y.underlying.symbol}`
+          ),
+      };
+    })
+    .sort((a, b) => a.apy - b.apy);
```
jeffywu, line 27: I think on all this dashboards lets sort by TVL descending (highest TVL first).

PR #638, opened 2024-02-27
feat: add filters for new mainnet proxy contract

sbuljac:


tenderly/actions/src/index.ts
```diff
@@ -138,6 +138,7 @@ export const accountsFn: ActionFn = async (
         'x-auth-token': dataServiceAuthToken
       },
       body: JSON.stringify({
+        network: context.metadata.getNetwork(),
```
jeffywu, line 141: I guess in the future this assumes that our network string will always match their network string? I think that is fine but there are some cases like ethers where mainnet is returned as "homstead" for the network name. Perhaps that will not be the case here.

jeffywu (approved):

tenderly/actions/src/index.ts
sbuljac: image
this is the `Network` type returned by `getNetwork`

PR #641, opened 2024-02-29
Feat/dashboard list view

matthew-garrett:
- Adds list view for dashboard pages
- Refactors dashboard data hooks


packages/common/mui/src/lib/product-dashboard/components/dashboard-card.tsx
```diff
@@ -161,10 +171,23 @@ const GridCard = styled(Box)(
...
+const GridCardTitle = styled(H4)(
+  ({ theme }) => `
+    font-family: Avenir Next;
     display: block;
     text-align: left;
+    margin-bottom: ${theme.spacing(0.5)};
+      `
+);
+
+const SectionTitle = styled(Box)(
```
jeffywu, line 183: Is there no theme style for this?

packages/core-entities/src/client/yield-registry-client.ts
```diff
@@ -237,7 +237,11 @@ export class YieldRegistryClient extends ClientRegistry<YieldData> {
         debt.totalAPY,
         leverageRatio
       ),
-      strategyAPY: yieldData.totalAPY,
+      interestAPY: this.calculateLeveragedAPY(
```
jeffywu, line 240: Can we rename this to organicAPY? I think that is more clear

packages/shared/trade/src/hooks/use-max-yield.ts
```diff
@@ -18,13 +18,6 @@ export const useMaxYield = (network: Network | undefined) => {
     .map((y) => {
```
jeffywu, line 18: Do we need this second map statement now?

packages/common/mui/src/lib/product-dashboard/components/dashboard-card.tsx
matthew-garrett: The problem is the card/landing page typography style bullshit. We were using that but it breaks a lot in the data table when I used that theme. It was easier to create the couple things we needed for the cards instead of fixing the table

jeffywu: ok sounds good

PR #654, opened 2024-03-07
Feat/landing product cards

matthew-garrett:


jeffywu (approved):

packages/common/icons/src/lib/token-icon/token-icon.tsx
```diff
@@ -473,7 +473,9 @@ export function TokenIcon({
     <Box position="relative">
       {networkIcon && size === 'medium' && (
         <img
-          src={networkIcon.img}
+          src={
```
jeffywu, line 476: Can we pull this out and just make it a variable?

PR #659, opened 2024-03-11
Feat/vault reinvestment table

matthew-garrett:


packages/features/vault/src/hooks/use-vault-reinvestment-table.tsx
```diff
@@ -0,0 +1,84 @@
...
+  let result = [] as any[];
+
+  if (
+    vaultAddress &&
+    reinvestmentData &&
+    reinvestmentData[vaultAddress] &&
+    network &&
+    deposit
+  ) {
+    result = reinvestmentData[vaultAddress].map((data) => {
```
jeffywu, line 59: Can you limit this to the 25 most recent values? So also sort this by timestamp descending so the most recent reinvestment is at the top.

PR #665, opened 2024-03-13
feat: calculate and save risk profiles

sbuljac:


apps/accounts/src/index.ts
```diff
@@ -10,10 +10,15 @@ export interface Env {
   NX_DD_API_KEY: string;
   SUPPORTED_NETWORKS: Network[];
   ACCOUNT_CACHE_R2: R2Bucket;
+  AUTH_KEY: string;
 }
 
 export default {
   async fetch(request: Request, env: Env): Promise<Response> {
+    const authKey = request.headers.get('x-auth-key');
```
jeffywu, line 18: I don't think we need to enforce an auth key on this

jeffywu (approved):

apps/accounts/src/index.ts
sbuljac: Somebody could just call it in a loop and incur unnecessary cost on Cloudflare, right?

PR #670, opened 2024-03-13
build: updates red usage

matthew-garrett:
- Portfolio Overview: Change red to "black" on debts column

- Portfolio Holdings: Change all negative numbers to "black", Change positive numbers in Total Earnings to green #1F9B99 including incentives tooltip numbers, Filter Borrows to bottom and add "Debt Positions" label

- Fixes edge case: Yield products with negative "Total Earnings" will be red

- Transaction History: Add icons, Change text to green for positive numbers, "black" for negative (Only underlying amount column)

- Transaction page: Change to Green/Black type for positive/negative numbers on Transaction Summary

- Confirmation screens: Change all values to "black"


packages/features/portfolio/src/containers/portfolio-holdings/use-portfolio-holdings.tsx
```diff
@@ -114,6 +115,44 @@ export function usePortfolioHoldings() {
     }
   }, [expandedRows, setExpandedRows, Columns]);
 
+  const portfolioHoldingsData =
+    toggleOption === 0 && !isBlocked && groupedRows.length > 0
+      ? groupedHoldings
+      : detailedHoldings;
+
+  function insertDebtDivider(arr) {
```
jeffywu, line 123: Can you move this function declaration outside of the hook? If it is declared inside it will be recreated on every render

jeffywu (approved):

PR #704, opened 2024-03-26
Bugfix/sort markets table

matthew-garrett:
- Fixes exist early copy
- Fixes dual incentive icons
- Adds organic bottom value on provide liquidity and leveraged liquidity grid cards
- Makes sure highest APY shows on lend fixed grid cards
- Fixes decimals on Risk Overview table
- Moves sDAI to stable coins instead of LSDs
- Fixes negative total earnings but on portfolio overview
- Sorts markets table by highest APY

### DIFF
--- a/packages/common/mui/src/lib/product-dashboard/components/dashboard-card.tsx
+++ b/packages/common/mui/src/lib/product-dashboard/components/dashboard-card.tsx
@@ -30,6 +30,9 @@ export const DashboardCard = ({
   const theme = useTheme();
   const hideFooter = !bottomValue && !incentiveSymbols;
 
+  console.log({ incentiveValue });
+  console.log({ incentiveSymbols });
+
   return (
     <GridCard onClick={() => routeCallback()}>
       <Box
@@ -123,15 +126,19 @@ export const DashboardCard = ({
                 }}
               />
             ))}
-            <Box
-              sx={{
-                color: theme.palette.typography.main,
-                marginRight: theme.spacing(0.5),
-              }}
-            >
-              {incentiveValue}
-            </Box>
-            <FormattedMessage defaultMessage={'Incentive APY'} />
+            {incentiveValue && (
+              <>
+                <Box
+                  sx={{
+                    color: theme.palette.typography.main,
+                    marginRight: theme.spacing(0.5),
+                  }}
+                >
+                  {incentiveValue}
+                </Box>
+                <FormattedMessage defaultMessage={'Incentive APY'} />
+              </>
+            )}
           </SectionTitle>
         )}
       </GridCardFooter>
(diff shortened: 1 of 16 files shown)


packages/common/mui/src/lib/product-dashboard/components/dashboard-card.tsx
```diff
@@ -123,15 +123,19 @@ export const DashboardCard = ({
                 }}
```
jeffywu, line 123: The issue here is that you need to check if there are NOTE incentives on the card or not. Some currencies no longer have NOTE incentives and therefore we should not show the NOTE icon.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-leveraged-grid.tsx
```diff
@@ -28,6 +31,47 @@ export const useLiquidityLeveragedGrid = (
         (n) => n.asset.balance.underlying.symbol === y.underlying.symbol
       );
 
+      const handleBottomValue = () => {
```
jeffywu: You shouldn't use inline functions like this, this creates a performance overhead. You should just put the logic inline without creating a function or make a function outside of the hook.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-leveraged-grid.tsx
```diff
@@ -66,12 +110,29 @@ export const useLiquidityLeveragedGrid = (
             ),
         incentiveSymbols: currentPosition
           ? getTotalIncentiveSymbol(
-              currentPosition?.asset.marketYield?.secondaryIncentives?.symbol,
-              currentPosition?.asset.marketYield?.noteIncentives?.symbol
+              currentPosition?.asset.marketYield?.secondaryIncentives
```
jeffywu: There is no difference between the `marketYield` object and the `y` object. You don't need to do a `?` on the currentPosition state here. You can just use `getTotalIncentiveSymbol(y?.....)` for everything.

Same for the if clause above.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-leveraged-grid.tsx
jeffywu: You do not need to use `marketYield` in here. You can use `y` regardless of currentPosition.

packages/core-entities/src/token-balance.ts
```diff
@@ -409,17 +409,38 @@ export class TokenBalance {
     decimalPlaces?: number,
     abbr = true,
     useThousandsAbbr = true,
-    locale = 'en-US'
+    locale = 'en-US',
+    hideSmallNegativeValues = false
   ) {
+    function containsNonZeroNumber(str: string) {
```
jeffywu: Move this function declaration to the utils package instead. Inline function declarations incur a performance overhead.

PR #720, opened 2024-04-01
Feat/react helmet

matthew-garrett:
- Adds react helmet and meta tags for dashboard, portfolio, and markets pages
- Fixes vault dashboard list view link bug

### DIFF
--- a/apps/web/src/containers/Markets/index.tsx
+++ b/apps/web/src/containers/Markets/index.tsx
@@ -10,9 +10,8 @@ import {
   useMarketTableDropdowns,
   useAllNetworksToggle,
 } from './hooks';
-import { FeatureLoader, MetaTagManager } from '@notional-finance/shared-web';
+import { FeatureLoader } from '@notional-finance/shared-web';
 import { MarketsMobileNav, MobileFilterOptions } from './components';
-import { META_TAG_CATEGORIES } from '@notional-finance/util';
 
 export const Markets = () => {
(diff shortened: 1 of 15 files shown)


apps/web/src/containers/Markets/index.tsx
```diff
@@ -34,6 +35,7 @@ export const Markets = () => {
 
   return (
     <FeatureLoader featureLoaded={true}>
+      <MetaTagManager metaTagCategory={META_TAG_CATEGORIES.MARKETS} />
```
jeffywu: Would it work better to put these meta tag manager components inside the `AppLayoutRoute` component and then have a dictionary of paths => META_TAG_CATEGORIES enums?

packages/shared/web/src/meta-tag-manager/meta-tag-manager.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+interface MetaTagManagerProps {
+  metaTagCategory: META_TAG_CATEGORIES;
+}
+
+export const MetaTagManager = ({ metaTagCategory }: MetaTagManagerProps) => {
+  return (
+    <Helmet>
+      <title>{metaTagData[metaTagCategory].title}</title>
```
jeffywu, line 61: This is a bit dangerous if `metaTagData[metaTagCategory]` ends up undefined, since you're using regex the type checking doesnt really work here.

PR #737, opened 2024-04-05
build: adds new network icons fixes bugs

matthew-garrett:
- Adds new network icons
- Fixes table spacing bug
- Fixes confirmation spacing bug


packages/shared/trade/src/transaction-confirmation/transaction-confirmation.tsx
```diff
@@ -69,7 +69,7 @@ export const TransactionConfirmation = ({
   }, []);
 
   return (
-    <Box sx={{ minHeight: theme.spacing(132) }}>
+    <Box sx={{ minHeight: theme.spacing(162) }}>
```
jeffywu, line 72: Won't this cause issues if the porfolio comparison is also expanded?

packages/shared/wallet/src/network-selector/network-selector.tsx
```diff
@@ -42,6 +39,8 @@ export const NetworkSelectorButton = ({
   balance,
 }: NetworkSelectorButtonProps) => {
   const theme = useTheme();
+  const networkIcon =
```
jeffywu: This isn't going to work if we expand to more networks, can we make a reusable Mui component for the network icons instead of sprinking the SVGs everywhere?

matthew-garrett: How about I update the TokenIcon component to take a ethNetwork and arbNetwork icon string and also update the getNetworkSymbol to return those strings so we can add a new network icon to the TokenIcon and getNetworkSymbol

PR #775, opened 2024-04-22
build: adds intercom update with arb and mainnet usd balance values

matthew-garrett:
- Adds intercom update call passing wallet address, arbitrum and mainnet use wallet balance values

### DIFF
--- a/apps/web/src/containers/App/App.tsx
+++ b/apps/web/src/containers/App/App.tsx
@@ -4,6 +4,7 @@ import {
   useSanctionsBlock,
   useWalletConnectedNetwork,
 } from '@notional-finance/notionable-hooks';
+import { IntercomProvider } from 'react-use-intercom';
 import {
   FeatureLoader,
   TrackingConsent,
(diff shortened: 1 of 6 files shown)


apps/web/src/containers/App/App.tsx
```diff
@@ -269,6 +269,7 @@ export const App = () => {
   const {
     state: { themeVariant },
   } = globalState;
+
   const notionalTheme = useNotionalTheme(themeVariant);
 
   return (
```
jeffywu, line 275: You don't need the Intercom Provider in here anymore?

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -0,0 +1,38 @@
...
+        );
+        networkBalanceData[network] = usdBalance
+          ? formatNumberAsAbbr(usdBalance)
+          : '';
+      });
+
+      update({
+        userId: selectedAccount,
+        name: selectedAccount,
+        customAttributes: {
```
jeffywu: You don't need to destructure here? Can't you just do `customAttributes: networkBalanceData`

apps/web/src/containers/App/App.tsx
matthew-garrett: Good call, it must have gotten washed out when I rebased

PR #781, opened 2024-04-29
build: adds portfolio link with view as address

matthew-garrett:
- Adds copy transaction data link in confirmation screen
- Makes the wallet connection module a fixed width
- Adds a banner to show when the subgraph is down


packages/shared/trade/src/transaction-confirmation/components/portfolio-compare.tsx
```diff
@@ -90,6 +106,23 @@ export const PortfolioCompare = ({
...
+          marginTop: theme.spacing(2),
+          cursor: 'pointer',
+        }}
+        onClick={handleCopy}
+      >
+        <FormattedMessage defaultMessage={'Copy Transaction Details'} />
+        <CopyIcon
+          fill={theme.palette.typography.accent}
+          sx={{
+            marginLeft: '0.25rem',
```
jeffywu: Can we switch these to theme.spacing and px? Trying to destroy references to rems

matthew-garrett: Sounds good

PR #789, opened 2024-05-06
feat: add worker for market initialization and account settlement

sbuljac:


apps/initialize-markets/jest.config.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable */
+export default {
+  displayName: 'rewards',
```
jeffywu: update display name

apps/initialize-markets/src/Markets.ts
```diff
@@ -0,0 +1,181 @@
...
+        // .then(r => {console.log(r); return r;})
+        .then((r: { data: { accounts: [{ id: string }] } }) =>
+          r.data.accounts.map((a) => a.id)
+        );
+      accounts.push(...tempAccounts);
+      skip += tempAccounts.length;
+    } while (tempAccounts.length === GRAPH_MAX_LIMIT);
+
+    console.log("accounts: ", accounts.length);
+    const perChunk = 100;
```
jeffywu: there is a helper function `batchArray` that you can use here

apps/initialize-markets/src/Markets.ts
```diff
@@ -0,0 +1,181 @@
...
+              vaultAddress
+              maturity
+            }
+          }
+        }`
+      }),
+    })
+      .then((r) => r.json())
+      // group accounts per vault
+      .then((r: { data: { balances: [{ account: { id: string }, token: { vaultAddress: string } }] } }) =>
```
jeffywu: there is a helper function `groupArrayToMap` or `groupArrayByKey` that you could use here.

package.json
```diff
@@ -39,6 +39,7 @@
     "cross-fetch": "^3.1.5",
     "dotenv": "^16.0.2",
     "ethers": "^5.7.1",
+    "ethers-v6": "npm:ethers@latest",
```
jeffywu: Is ethers-v6 strictly required? Would like to keep dependencies clean but we can use it if you need to

jeffywu (approved):

apps/initialize-markets/src/Markets.ts
```diff
@@ -0,0 +1,164 @@
...
+  "function initializeAllMarkets()",
+  "function checkInitializeAllMarkets() view returns (bool canExec, bytes memory execPayload)",
+];
+
+const SETTLE_ACCOUNTS_ABI = [
+  "function settleAccounts(address[] calldata accounts)",
+  "function settleVaultsAccounts((address vaultAddress,address[] accounts)[])",
+];
+
+const GRAPH_MAX_LIMIT = 100;
```
jeffywu: This can go up to 1000 I believe.

PR #790, opened 2024-05-06
Bugfix/trade summary earnings

matthew-garrett:


packages/shared/notionable-hooks/src/summary/use-trade-summary.tsx
```diff
@@ -731,5 +731,32 @@ export function useTradeSummary(state: VaultTradeState | TradeState) {
...
+        {
+          displayValue: `+${earnings?.toDisplayStringWithSymbol()}`,
+          isNegative: false,
+          showPositiveAsGreen: true,
+        },
+      ],
+    },
+  };
+
+  // TODO: Add all of the borrow and repay debt shit
```
jeffywu: Can we fix the TODO thing? I don't think the tradeType restriction is required?

packages/shared/notionable-hooks/src/summary/index.ts
```diff
@@ -35,27 +41,27 @@ export const OrderDetailLabels = defineMessages({
 
 export const TradeSummaryLabels = {
```
jeffywu: Is there a reason why these changed? They are used in many places, not sure if I changed them or you did...

packages/common/mui/src/lib/data-table-cells/multi-value-cell/multi-value-cell.tsx
```diff
@@ -47,7 +54,8 @@ export const MultiValueCell = ({ cell, row, column }): JSX.Element => {
                     sx={{
                       marginBottom: '0px',
                       width: '100%',
-                      fontWeight: row.original.currency === 'Total' ? 600 : 500,
+                      fontWeight:
+                        isTotalRow || row.original.isEarningsRow ? 600 : 500,
                       color:
```
jeffywu, line 58: Earnings font size is wrong:
image

packages/shared/notionable-hooks/src/summary/index.ts
jeffywu: Oh I see, I'd rather use the ones that I have than switch them to what Isaac has in the mocks. Can you revert these changes here?

matthew-garrett: I made these changes in accordance with what was in the new mocks for trade summary table

packages/shared/notionable-hooks/src/summary/use-trade-summary.tsx
```diff
@@ -731,5 +731,37 @@ export function useTradeSummary(state: VaultTradeState | TradeState) {
...
+        {
+          displayValue: `${prefixSymbol}${earnings?.toDisplayStringWithSymbol()}`,
+          isNegative: false,
+          showPositiveAsGreen: !isDebt,
+        },
+      ],
+    },
+  };
+
+  if (
```
jeffywu: Why is this if statement so involved? Shouldn't it just check if the earnings is undefined or not? In fact, why not just put this whole code block you added inside an:

```
if (earnings) {
...
}
```

matthew-garrett: It is involved because just earnings wasn't enough to prevent a undefined value from being shown. We never want to show the earnings row on any of these pages so this was an effective way of preventing that.

matthew-garrett: Would this be better 

` if (earnings && earnings.isPositive()) {`

PR #824, opened 2024-05-16
Feat/note page upgrades

matthew-garrett:
- Adds real governance data
- Adds delegate data
- Moves note supply and staked note data calls to notionable
- Adds Time Frame options
- Updates note supply title value on hover
- Other style fixes


apps/web/src/containers/NoteView/note-governance/note-governance.tsx
```diff
@@ -13,16 +13,31 @@ import { FormattedMessage } from 'react-intl';
...
+  active: boolean;
   theme: NotionalTheme;
 }
 
 export const NoteGovernance = () => {
   const theme = useTheme();
   const govData = useGovernanceData();
 
+  const calculateDaysDifference = (timestamp: number): number => {
+    const secondsPerDay = 24 * 60 * 60;
```
jeffywu: There is already a constant SECONDS_IN_DAY

apps/web/src/containers/NoteView/note-governance/note-governance.tsx
```diff
@@ -13,16 +13,31 @@ import { FormattedMessage } from 'react-intl';
...
+    const currentDate = Math.floor(Date.now() / 1000);
```
jeffywu: There is already a function `getNowSeconds()`

apps/web/src/containers/NoteView/note-governance/note-governance.tsx
```diff
@@ -13,16 +13,31 @@ import { FormattedMessage } from 'react-intl';
...
-  open: boolean;
+  active: boolean;
   theme: NotionalTheme;
 }
 
 export const NoteGovernance = () => {
   const theme = useTheme();
   const govData = useGovernanceData();
 
+  const calculateDaysDifference = (timestamp: number): number => {
```
jeffywu: Move this function to the utils package.

apps/web/src/containers/NoteView/note-governance/use-governance-data.tsx
```diff
@@ -1,15 +1,54 @@
...
-      title:
-        '[NIP-60] - Increase USDT/USDC.e Leveraged Vault Capacity (Arbitrum)',
-    },
-    {
-      title: '[NIP-62] - List a Convex crvUSD/USDT leveraged vault (Arbitrum)',
-    },
-  ];
+  const [govData, setGovData] = useState<any[]>([]);
+
+  if (govData.length > 0) return govData; // Return if data is already present
```
jeffywu: Shouldn't this be inside a useEffect?

apps/web/src/containers/NoteView/note-supply/note-summary.tsx
```diff
@@ -16,23 +16,38 @@ import {
   H5,
   NoteChart,
   H2,
+  DateRangeButtons,
+  dateRangeData,
```
jeffywu: This shouldn't be in the MUI package in my opinion. I would put it somewhere inside the `NoteView` container.  Also it should be named something more descriptive like `ValidDateRanges`.

packages/common/mui/src/lib/date-range-buttons/date-range-buttons.tsx
```diff
@@ -0,0 +1,81 @@
...
+interface DateRangeButtonsProps {
+  setDateRange: Dispatch<SetStateAction<number>>;
+  dateRange: number;
+}
+interface LabelButtonProps {
+  active?: boolean;
+  theme: NotionalTheme;
+}
+
+export const dateRangeData = [
```
jeffywu: Rename this to `ValidDateRanges`

packages/common/mui/src/lib/date-range-buttons/date-range-buttons.tsx
```diff
@@ -0,0 +1,81 @@
...
+  { displayValue: '30d', value: 2592000 },
```
jeffywu: Use constant math instead here:

`30 * SECONDS_IN_DAY`
`90 * SECONDS_IN_DAY`
`SECONDS_IN_YEAR_ACTUAL`
`2 * SECONDS_IN_YEAR_ACTUAL`

apps/web/src/containers/NoteView/staked-note/staked-note.tsx
```diff
@@ -140,15 +151,15 @@ export const StakedNote = () => {
               size="large"
               sx={{
                 fontFamily: 'Avenir Next',
```
jeffywu: You can revert this disabling of the stake note thing, I will merge a PR after you to get this to work. It will sit at the `/stake` route.

packages/common/mui/src/lib/note-chart/note-chart.tsx
```diff
@@ -1,10 +1,14 @@
 import type { CSSProperties } from 'react';
 import { Box, styled, useTheme } from '@mui/material';
-import type { EChartsOption, SetOptionOpts } from 'echarts';
+// import { use, type EChartsOption, type SetOptionOpts } from 'echarts';
```
jeffywu: Remove commented line?

packages/common/mui/src/lib/note-chart/note-chart.tsx
```diff
@@ -137,16 +156,26 @@ export const NoteChart = ({
...
-        return `<div style="background-color: #041D2E; width: 100px; box-shadow: -2px 1px 24px 0px rgba(135, 155, 215, 0.20), 0px 4px 16px 0px rgba(121, 209, 212, 0.40); padding: 8px; height: 58px;">
-                  <div style="color: #BCD4DB; font-family: Avenir Next; font-weight: 500;">${params[0].axisValue}</div>
-                  <div style="color: #ffffff; font-family: Avenir Next; font-weight: 500;" >${params[0].data}</div>
+        const value = formatToolTipValueAsFiat
+          ? formatNumberAsAbbr(params[0].data[1], 4, baseCurrency)
+          : formatNumber(params[0].data[1]);
+        if (noteNumCallback) {
+          noteNumCallback(params[0].data[1]);
+        }
+        return `<div style="background-color: #041D2E; width: fit-content; box-shadow: -2px 1px 24px 0px rgba(135, 155, 215, 0.20), 0px 4px 16px 0px rgba(121, 209, 212, 0.40); padding: 8px; height: 58px;">
```
jeffywu, line 164: Can we make a styled component instead of embedding all this CSS like this?

packages/shared/helpers/src/number-helpers.ts
```diff
@@ -71,9 +72,10 @@ export function formatNumberAsAbbr(
     suffix = 'b';
     num = num / 1_000_000_000;
   }
+  
 
   const symbol =
-    baseCurrency && FiatSymbols[baseCurrency] ? FiatSymbols[baseCurrency] : '$';
+    baseCurrency && FiatSymbols[baseCurrency] ? FiatSymbols[baseCurrency] : hideSymbol ? '' : '$';
```
jeffywu: Shouldn't the `hideSymbol` boolean come first in the ternery? If the initial condition is met then it won't work?

apps/web/src/containers/NoteView/note-supply/note-summary.tsx
matthew-garrett: I put it in mui since I figured we would probably want to use is with other charts in the app. Where do you purpose I put it instead? Shared web?

packages/shared/helpers/src/number-helpers.ts
matthew-garrett: That is not the case. The first condition is not met and it moves on to hideSymbol.

matthew-garrett: I can reorder it if you want

packages/common/mui/src/lib/note-chart/note-chart.tsx
matthew-garrett: Thought that was something we could do when we extend the chart in the future. We could have a number of different tooltip components like our table cell components to handle different situations. I just didn't take the time to do that right now.

packages/shared/helpers/src/number-helpers.ts
matthew-garrett: After further investigation into the docs I don't think when can use a normal component here. A string template has to be returned. I don't think it handle JSX

packages/common/mui/src/lib/note-chart/note-chart.tsx
matthew-garrett: After further investigation into the docs I don't think when can use a normal component here. A string template has to be returned. I don't think it handle JSX

jeffywu: ok sounds good

packages/util/src/time.ts
```diff
@@ -105,3 +111,12 @@ export const formatMaturity = (ts: number) => {
 export const floorToMidnight = (ts: number, offset = 0) => {
   return ts - offset - ((ts - offset) % SECONDS_IN_DAY);
 };
+
+export const getDaysDifference = (timestamp: number): number => {
+  const currentDate = Math.floor(Date.now() / 1000);
```
jeffywu: This should be getNowSeconds()

PR #832, opened 2024-05-17
Feat/intercom value updates

matthew-garrett:
- Adds new intercom values
- Fixes SideDrawer style issues when a intercom banner is present
- Fixes spacing issues in cookie message
- Fixes dead links

### DIFF
--- a/packages/shared/wallet/src/hooks/use-intercom-update.tsx
+++ b/packages/shared/wallet/src/hooks/use-intercom-update.tsx
@@ -14,21 +14,42 @@ export const useIntercomUpdate = () => {
 
   useEffect(() => {
     if (selectedAccount && !isReadOnlyAddress) {
-      const networkBalanceData = {};
+      const balanceData = {};
       SupportedNetworks.forEach((network) => {
         const account = networkAccounts ? networkAccounts[network] : undefined;
-        const usdBalance = account?.accountDefinition?.balances
-          .filter((t) => t.symbol !== 'sNOTE')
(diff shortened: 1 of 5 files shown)


packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -14,21 +14,42 @@ export const useIntercomUpdate = () => {
...
+          .reduce((acc, b) => acc + b.toFiat('USD').toFloat(), 0);
+
+        balanceData[network] = {
+          walletBalance: walletBalance || 0,
+          notionalBalance: notionalBalance || 0,
+        };
       });
 
+      const totalWalletBalance =
+        balanceData['mainnet'].walletBalance +
```
jeffywu: You should loop over SupportedNetworks instead

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -14,21 +14,42 @@ export const useIntercomUpdate = () => {
...
+        balanceData['mainnet'].notionalBalance +
+        balanceData['arbitrum'].notionalBalance;
+
+      const totalBalance = totalWalletBalance + totalNotionalBalance;
+
       update({
         userId: selectedAccount,
         name: selectedAccount,
-        customAttributes: networkBalanceData,
+        customAttributes: {
```
jeffywu, line 47: Don't we want these just as numbers instead? Intercom is not parsing the strings into numbers

PR #838, opened 2024-05-22
feat: use wrapped flash providers

sbuljac:


apps/liquidation-bot/src/NotionalV3Liquidator.ts
```diff
@@ -11,15 +11,14 @@ import {
 } from './types';
 import LiquidationHelper from './LiquidationHelper';
 import ProfitCalculator from './ProfitCalculator';
-import AaveFlashLoanProvider from './lenders/AaveFlashLender';
```
jeffywu, line 14: @sbuljac Can you split the changes to `liquidation-bot` into a separate pull request? I'm going to need to test them separately and will merge the `vault-liquidator` changes first.

PR #843, opened 2024-05-23
Feat/cool down

matthew-garrett:
- Adds cool down dates, countdown, and buttons
- Updates portfolio top margins
- Added the ability to pass custom links and TradeActionButton text props
- Adds pending txn boxes to cool down side bar and cool down drawer
- Adds switch network message to cool down drawer
- Adds secondary option for NOTE to the transaction approval flow


apps/web/src/containers/StakeNOTE/sidebars/cooldown.tsx
```diff
@@ -0,0 +1,142 @@
...
+  useSideDrawerManager,
+  useSideDrawerState,
+} from '@notional-finance/side-drawer';
+import { SNOTEWeightedPool } from '@notional-finance/core-entities';
+
+export const CoolDown = () => {
+  const theme = useTheme();
+  const context = useContext(NOTEContext);
+  const { sideDrawerOpen } = useSideDrawerState();
+  const { isReadOnlyAddress, onSubmit } = useTransactionStatus(Network.mainnet);
```
jeffywu: Can you show this component when a transaction is in progress?

image


Same thing on the StartCoolDown component in the portfolio

apps/web/src/containers/StakeNOTE/sidebars/cooldown.tsx
```diff
@@ -0,0 +1,142 @@
...
+            {stakeNoteStatus?.redeemWindowEnd
+              ? getDateString(stakeNoteStatus.redeemWindowEnd, {
+                  showTime: true,
+                  slashesFormat: true,
+                })
+              : '-'}
+          </LabelValue>
+        </Box>
+      </Box>
+      {isReadOnlyAddress || !account ? (
```
jeffywu, line 133: Should we make this a shared component?

matthew-garrett: It's not worth it imo. Only the date portion would be reusable and they're only used here and in the unstake drawer in portfolio. If the countdown was the same and the container styling was the same I think it would be worth it.

jeffywu: Ok

apps/web/src/containers/StakeNOTE/sidebars/cooldown.tsx
matthew-garrett: I am trying to stake note to be able to actually test this but when I try to stake some of my note I get a gas limit error. Trying to find a work around

jeffywu: Is it not triggering the token approval setting?

matthew-garrett: I haven't been able to get the stake note flow to work. So I haven't been able to get to this step with a actual wallet that is able to trigger the transaction

PR #854, opened 2024-05-29
build: merge v3/prod

jeffywu:


packages/shared/helpers/src/number-helpers.ts
```diff
@@ -56,26 +56,33 @@ export function formatNumberAsAbbr(
   decimalPlaces?: number,
   baseCurrency?: string,
   hideSymbol?: boolean,
```
jeffywu: @matthew-garrett can you change this to use `opt?: { hideSymbol?: boolean, removeKAbbr?: boolean }` as the parameter instead? You can push into this branch

PR #889, opened 2024-06-10
Feat/nav updates

matthew-garrett:
- Adds Leveraged Farming dashboards
- Adds leverage dropdown
- Removes Markets nav option
- Updates dropdown nav options
- Updates analytics dropdown
- Adds disabled state zero card state
- Updates tokens that are shown on state zero


packages/common/mui/src/lib/product-dashboard/product-dashboard.tsx
```diff
@@ -21,6 +21,7 @@ export interface DashboardDataProps {
   bottomLeftValue?: string | ReactNode;
   bottomRightValue?: string | ReactNode;
   hasPosition?: boolean;
+  isPointsVault?: boolean;
```
jeffywu: We will likely be extending the vault types in the future, can we make this an enum instead?
"Leveraged Yield Farming"
"Leveraged Points Farming"

etc.

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
```diff
@@ -10,18 +10,27 @@ import {
...
   const selectedNetwork = useSelectedNetwork();
 
+  const leveragedLiquidityData = activeTokenData?.find(
+    (data) => data.product === 'Leveraged Liquidity'
+  );
+  const leveragedYieldFarming = activeTokenData?.find(
+    (data) => data.product === 'Leveraged Vault'
+  );
+  const leveragedPointsFarming = activeTokenData?.find(
+    (data) => data.product === 'Leveraged Vault' && data.pointMultiples
```
jeffywu, line 31: I wonder if we should change `data.product` instead of having this pointMultiples logic here. Is that going to be a big refactor?

Having data.product be "Leveraged Yield Farming" or "Leveraged Point Farming"

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
```diff
@@ -95,16 +104,26 @@ export const useCardData = (
 
   const leveragedData = [
     {
```
jeffywu: Are the three cards pretty similar? Can you make each card a component that takes the `YieldData` object and some other stuff as props? The code will look cleaner.

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
```diff
@@ -95,16 +104,26 @@ export const useCardData = (
...
-      accentTitle: <FormattedMessage defaultMessage={'Incentivized Yield'} />,
+      accentTitle: <FormattedMessage defaultMessage={'Incentivised Yield'} />,
```
jeffywu: `Incentivised` with an "s" is the British spelling and we won that war

packages/shared/web/src/dashboard-view/hooks/use-dashboard-config.tsx
```diff
@@ -9,11 +9,49 @@ export const config = {
...
     containerData: {
       heading: defineMessage({
-        defaultMessage: 'Leveraged Vaults',
+        defaultMessage: 'Leveraged Points Farming',
         description: 'page heading',
       }),
       subtitle: defineMessage({
-        defaultMessage: `Get one-click access to sophisticated DeFi yield strategies and dial up your leverage for maximum efficiency.`,
+        defaultMessage: `Boost your points yield from exciting partner protocols using leverage.`,
+        description: 'page heading subtitle',
```
jeffywu: Is this in here twice?

packages/shared/web/src/dashboard-view/hooks/use-leverage-points-grid.tsx
```diff
@@ -0,0 +1,137 @@
...
+} from '@notional-finance/notionable-hooks';
+import { useHistory } from 'react-router';
+import { DashboardGridProps, DashboardDataProps } from '@notional-finance/mui';
+import { Network, PRODUCTS } from '@notional-finance/util';
+import { formatNumberAsAbbr } from '@notional-finance/helpers';
+import { defineMessage } from 'react-intl';
+import { PointsIcon } from '@notional-finance/icons';
+import { Box } from '@mui/material';
+
+export const useLeveragePointsGrid = (network: Network): DashboardGridProps => {
```
jeffywu: It seems like the grid hook and the list hook are pretty similar, lots of repeated code. Is there a way to make this more DRY?

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
matthew-garrett: The product here is coming from the allYields object in global state. Updating that would break a lot of stuff.

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
matthew-garrett: I like handling it with this array of objects because it makes it easy to update the values in the future. If I pass down the yieldData, formatting all of this data is still going to have to happen just at the card level.

packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
matthew-garrett: vs code thinks the z version is a error

PR #899, opened 2024-06-14
build: adds spindl

matthew-garrett:


packages/shared/wallet/src/hooks/use-connect.ts
```diff
@@ -34,7 +35,10 @@ export const useConnect = () => {
     (walletLabel?: string) => {
       // No change to wallets, nothing to do here.
       if (!walletLabel || currentLabel === walletLabel) return;
-      connect({ autoSelect: { label: walletLabel, disableModals: true } });
+      connect({ autoSelect: { label: walletLabel, disableModals: true } }).then((resp) => {
+        // NOTE: this handles initial wallet connection
```
jeffywu, line 39: This will fire multiple times unless it goes into the observables. Same goes for the useEffect.

https://github.com/notional-finance/notional-monorepo/blob/be08117fb2910814092edc1abc8852a384296e60/packages/notionable/src/global/sagas/on-wallet-connect.ts

These should get fired off here:
image

apps/web/src/containers/App/App.tsx
jeffywu, line 1: Can you rebase this onto v3/prod instead?

packages/shared/wallet/src/hooks/use-connect.ts
matthew-garrett: I tested both and the spindl call in in the .then only gets called once. The spindl call in the useEffect get's hit 3 times when a user switches there wallet, which is not ideal but not terrible (I thought). I'll move them to that observable

PR #906, opened 2024-06-14
build: disables routes adds deprecation banner

matthew-garrett:


packages/features/portfolio/src/components/deprecation-message/deprecation-message.tsx
```diff
@@ -0,0 +1,80 @@
...
+              ),
+            }}
+          />
+        </Body>
+      </Box>
+      <ButtonBox>
+        <Button
+          variant="outlined"
+          size="large"
+          href="https://blog.notional.finance/notional-v2-deprecation-plan/"
```
jeffywu: Ok so we just need to replace this URL right?

jeffywu (approved):

packages/features/portfolio/src/components/deprecation-message/deprecation-message.tsx
matthew-garrett: Yes, we need to replace that with the new one whenever Teddy publishes it

PR #908, opened 2024-06-16
fix: properly use wait function

sbuljac:


jeffywu (approved):

jeffywu:
@sbuljac can you also add this to the deployWorkers.sh file?

PR #909, opened 2024-06-16
Feat/data service secrets manager

sbuljac:


jeffywu (approved):

gcp/tx-relay/src/Signer.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  // in order to properly fetch correct nonce for tx we need to throttle here per key
+  await throttle(keysToUse[to]);
+
+  const provider = ethers.getDefaultProvider(rpcUrls[network]);
+
+  const signer = new GcpKmsSigner({ ...kmsCredentials, keyId: keysToUse[to] }).connect(provider);
+
+  const tx = await signer.sendTransaction({
```
jeffywu: @sbuljac if this throws an error can we log a special transaction failed event into datadog? I will set up a monitor to alert on that

PR #958, opened 2024-07-01
Fix/enter pool single sided

sbuljac:


apps/rewards/src/vaults.ts
```diff
@@ -71,57 +72,57 @@ export const vaults: Partial<Record<Network, Array<Vault>>> = {
...
       poolTokens: ['0x596192bB6e41802428Ac943D2f1476C1Af25CC0E', ezETH, ETH],
-      tokenWeights: [0, 50, 50],
+      tokenWeights: [0, 0, 100],
     },
     {
       address: '0x32D82A1C8618c7Be7Fe85B2F1C44357A871d52D1',
       rewardTokens: [BAL, AURA],
       poolTokens: ['0x05ff47AFADa98a98982113758878F9A8B9FddA0a', rETH, weETH],
-      tokenWeights: [0, 50, 50],
+      tokenWeights: [0, 0, 100],
```
jeffywu, line 105: Isn't rETH more liquid?

sbuljac: no

sbuljac: yest it is, my brain read wsteth instead of weEth

PR #994, opened 2024-07-19
feat: use v2 0x api

sbuljac:


apps/note-reward-manager/src/treasury_manager.ts
```diff
@@ -129,6 +130,7 @@ export default class TreasuryManager {
             sellToken: this.COMP,
             buyToken: this.WETH,
             sellAmount,
+            taker: this.proxy.address,
```
jeffywu, line 133: We will also need to add the chainId and taker address the the 0x api calls in the liquidation bot

sbuljac: ah, right

sbuljac: I updated the bot but did not test it

PR #1027, opened 2024-08-05
Feat/vault reinvestment date

matthew-garrett:


packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -224,6 +235,20 @@ export const useVaultHoldingsTable = () => {
       const totalPoints =
         arbPoints?.find(({ token }) => token === v.vaultShares.tokenId)
           ?.points || 0;
+      const vaultReinvestmentData = reinvestmentData?.[v.vaultAddress];
```
jeffywu: Do a check in case the vault address is not in the object or else this will crash

packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -224,6 +235,20 @@ export const useVaultHoldingsTable = () => {
...
+      let arbReinvestmentDate = '';
+      let mainnetReinvestmentDate = '';
+
+      if (vaultReinvestmentData) {
+        arbReinvestmentDate = moment(vaultReinvestmentData[0].timestamp * 1000)
```
jeffywu: Do a length check or else this will crash

packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -224,6 +235,20 @@ export const useVaultHoldingsTable = () => {
       const totalPoints =
         arbPoints?.find(({ token }) => token === v.vaultShares.tokenId)
           ?.points || 0;
+      const vaultReinvestmentData = reinvestmentData?.[v.vaultAddress];
+      let arbReinvestmentDate = '';
+      let mainnetReinvestmentDate = '';
+
+      if (vaultReinvestmentData) {
```
jeffywu: Move the code inside this if block into a separate helper function so we can easily update the reinvestment cadences.

packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -252,6 +277,18 @@ export const useVaultHoldingsTable = () => {
         },
       ];
 
+      if (!points) {
+        subRowData.push({
+          label: (
+            <FormattedMessage defaultMessage={'Time to Next Reinvestment'} />
+          ),
+          value:
+            network === 'mainnet'
```
jeffywu: Don't use string comparison here, use Network.mainnet instead, move this into the helper function as well

matthew-garrett: What're you wanting to move to a helper? This if and the else if subRowData stuff?

PR #1028, opened 2024-08-06
Deprecate Notional V2 Site

matthew-garrett:
- Removes initializeNotional calls 
- Removes footer data
- Adds lending message


apps/web/src/containers/App/index.tsx
```diff
@@ -79,20 +72,10 @@ export const App = () => {
     }
   }, []);
 
-  useEffect(() => {
-    initPlausible();
-    trackPageview();
-  }, [trackPageview]);
-
   useEffect(() => {
     initApplication();
```
jeffywu, line 76: What does this call do? Do we need to remove it as well?

matthew-garrett: It calls the above initApplication that sets up onboard so we still need it

PR #1030, opened 2024-08-08
Feat/analytics dropdown

matthew-garrett:
- Adds markets dropdown
- Adds network to markets route
- Adds accounts menu for all accounts and vault accounts


gcp/analytics/src/utils/notional-utils.ts
```diff
@@ -0,0 +1,4 @@
+export const capitalizeFirstLetter = (string) => {
```
jeffywu: Isn't there a CSS property for this?

packages/common/icons/src/utils.tsx
```diff
@@ -0,0 +1,17 @@
+export enum Network {
```
jeffywu: I'm not sure this is a good idea...we should try to actually fix the import issue.

gcp/analytics/src/layout/DashboardLayout/Header/HeaderContent/index.tsx
```diff
@@ -48,7 +49,25 @@ export default function Header() {
     }
   }, [pathname, setPathname]);
 
-  const availableTokens = ['ETH', 'DAI', 'USDC', 'GHO'];
+  const mainnetTokens: string[] = ['USDC', 'rETH', 'USDT', 'ETH', 'wstETH', 'DAI', 'sDAI', 'ARB', 'GHO', 'wBTC', 'FRAX'];
```
jeffywu: Can we move this off into a hook? We will need to fetch this list from an API.

gcp/analytics/src/layout/DashboardLayout/Header/HeaderContent/index.tsx
```diff
@@ -225,28 +280,79 @@ export default function Header() {
...
-          }}
-        >
-          {availableTokens.map((token, i) => (
-            <NextLink href={`/markets/${token}/overview`} passHref legacyBehavior key={i}>
-              <Link className="header-link" color="black" underline="none" sx={{ cursor: 'pointer' }}>
-                {token}
-              </Link>
-            </NextLink>
-          ))}
+        <Box sx={{ width: '767px', height: '100%', display: 'flex', justifyContent: 'space-between' }}>
```
jeffywu: We're using a lot of hardcoded `px` values in here, is that the convention for this project or do they have some sort of spacing method they use from the theme?

packages/common/icons/src/utils.tsx
matthew-garrett: It was being a pain in the ass and didn't seem worth it to me for one function and one enum

PR #1044, opened 2024-08-15
build: adds mst app state with baseCurrency and themeVariant

matthew-garrett:


packages/notionable/src/models/app-state.ts
```diff
@@ -0,0 +1,42 @@
...
+        setBaseCurrency(currency: FiatKeys) {
+            self.baseCurrency = currency;
+        },
+        setThemeVariant(variant: THEME_VARIANTS) {
+            self.themeVariant = variant;
+        },
+
+    }));
+
+    const AppStore = AppState.create({
```
jeffywu: Indentation looks wrong here

packages/notionable/src/models/app-state.ts
```diff
@@ -0,0 +1,42 @@
...
+        baseCurrency: userSettings?.baseCurrency ? userSettings?.baseCurrency : 'USD',
+        themeVariant: userSettings?.themeVariant
+        ? userSettings?.themeVariant
+        : THEME_VARIANTS.LIGHT,
+    });
+
+    export const useAppState = () => {
```
jeffywu: I think we should move this hook into notionable-hooks, the notionable package doesn't build cleanly with React

PR #1049, opened 2024-08-16
Feat/hero stats mobx

matthew-garrett:


packages/notionable/src/models/app-state.ts
```diff
@@ -1,16 +1,37 @@
...
+const HeroStats = types
+  .model('HeroStats', {
+    totalAccounts: types.number,
+    totalDeposits: types.number,
+    totalOpenDebt: types.number,
+  })
+  .actions((self) => ({
+    fetchKpiData: flow(function* () {
+      try {
+        const response = yield fetch('https://data-dev.notional.finance/kpi');
```
jeffywu: The hostname should be loaded from an environment variable

PR #1062, opened 2024-08-20
Feat/arb points mobx

matthew-garrett:
- Adds points store mobx model
- Adds init portfolio and init points dashboard files to setup react contexts for each feature. In the future we can create specific mobx models for each feature and the arb points data will just be a part of it. 
- Initializes the pointStore data from the wallet selector
- Wraps components that use pointStore data in mobx state tree observer HOC
- Fixes index related errors


apps/web/src/containers/TradingContest/points-dashboard/my-breakdown.tsx
```diff
@@ -3,15 +3,19 @@ import { THEME_VARIANTS } from '@notional-finance/util';
 import { useMyBreakdownTable } from './hooks';
 import { useNotionalTheme } from '@notional-finance/styles';
 import { ThemeProvider } from '@mui/material';
+import { observer } from 'mobx-react-lite';
+import { usePointsDashboardStore } from './init-points-dashboard';
 
 export const MyBreakdown = () => {
```
jeffywu: When you export the default that is wrapped in the `observer` HOC, we should remove the export here so that we don't accidentally import the wrong one function

packages/features/portfolio/src/components/side-nav/side-nav-options.tsx
```diff
@@ -29,9 +29,9 @@ export const SideNavOptons = ({ open }: SideNavOptonsProps) => {
 
   return (
     <Box>
-      {sideNavOptions.map(({ id, Icon, notifications, to }) => {
+      {sideNavOptions.map(({ id, Icon, notifications, to }, index) => {
         return (
-          <Box>
+          <Box key={index}>
```
jeffywu: We should use the `id` instead for the key.

apps/web/src/containers/TradingContest/points-dashboard/points-dashboard.tsx
```diff
@@ -16,11 +16,13 @@ import backgroundColors from '../assets/color-blobs.png';
 import { YourPointsOverview } from './your-points-overview';
 import { PointsDashboardManager } from './points-dashboard-manager';
 import { FormattedMessage } from 'react-intl';
-import { useYourPointsOverviewTables } from './hooks';
+import { usePointsDashboardStore } from './init-points-dashboard';
+import { observer } from 'mobx-react-lite';
 
 export const PointsDashboard = () => {
```
jeffywu: Same here, remove this export

packages/features/portfolio/src/containers/portfolio-holdings/portfolio-holdings.tsx
```diff
@@ -11,6 +11,7 @@ import {
   useAccountDefinition,
   useSelectedNetwork,
 } from '@notional-finance/notionable-hooks';
+import { observer } from 'mobx-react-lite';
 
 export const PortfolioHoldings = () => {
```
jeffywu: Remove export

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -59,22 +59,25 @@ export const useTotalsChart = () => {
...
-          ? colors.turquoise
-          : colors.neonTurquoise,
-      radius: [8, 8, 0, 0],
-      currencySymbol: FiatSymbols[baseCurrency]
-        ? FiatSymbols[baseCurrency]
-        : '$',
-      value: netWorth.toDisplayStringWithSymbol(2, true, false),
-    },
-  ];
+  const barConfig: BarConfigProps[] = useMemo(
```
jeffywu, line 62: Why do we have to memo this?

packages/notionable/src/models/points-store.ts
```diff
@@ -0,0 +1,47 @@
...
+          (t, { points }) => t + points,
+          0
+        );
+      } catch (error) {
+        console.warn('Error fetching arb points', error);
+      }
+      return self.arbPoints;
+    });
+
+    const initialize = (selectedAddress: string) => {
```
jeffywu, line 39: Can't you use the `afterCreate` hook here as well?

packages/shared/wallet/src/wallet-selector/wallet-selector.tsx
```diff
@@ -55,12 +56,19 @@ export function WalletSelector() {
     }
   };
 
+  useEffect(() => {
```
jeffywu, line 58: I wonder if this is the right place to put these intervals

packages/notionable/src/models/points-store.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (selectedAddress) {
+        fetchPoints(selectedAddress);
+      }
```
jeffywu, line 44: In the else condition you will need to set the values back to the initial values.

jeffywu: Add an action to clear the store back to the initial defaults.

packages/shared/wallet/src/wallet-selector/wallet-selector.tsx
jeffywu: Move this stuff into the use-connect hook

packages/notionable/src/global/global-state.ts
```diff
@@ -126,13 +126,6 @@ interface AddressState {
 
   /** Every supported network has an account object written to the state */
   networkAccounts?: Record<Network, AccountState>;
-  arbPoints?: {
```
jeffywu, line 129: Let's pull this out and make it a type so that we can re-use it for the props.

packages/shared/wallet/src/wallet-selector/wallet-selector.tsx
matthew-garrett: I think this is fine tbh

packages/notionable/src/models/points-store.ts
matthew-garrett: I don't think we want to use this here since we only want to initialize this if there is a address available right?

PR #1074, opened 2024-08-27
Feat/wallet mobx

matthew-garrett:
- Creates global AppContext, App Store, and useAppStore context hook
- Creates Wallet model and adds it to the App Store
- Updates many hooks relating to these changes


packages/shared/web/src/header/mobile-navigation/mobile-navigation.tsx
```diff
@@ -82,7 +80,7 @@ export function MobileNavigation({ ...rest }: TabsProps) {
   return window.innerWidth <= theme.breakpoints.values.sm ? (
     <>
       <Box sx={{ display: 'flex', justifyContent: 'end' }}>
-        {!wallet?.selectedAddress && isAccountPending && pathname !== '/' && (
+        {!wallet.userWallet?.selectedAddress && pathname !== '/' && (
```
jeffywu, line 84: Should we keep an isAccountPending flag here still? we can just stub one out instead of just deleting it?

packages/shared/web/src/header/use-leverage-earn-links.tsx
```diff
@@ -30,7 +30,7 @@ export const usePendingValues = (value: number | undefined) => {
 
 export const useLeverageEarnLinks = () => {
   const theme = useTheme();
-  const selectedNetwork = useDefaultNetwork();
+  const selectedNetwork = useSelectedNetwork();
```
jeffywu, line 33: Does this work on the landing page? There is no selected network there.

apps/deployWorkers.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -e
 
-yarn nx affected --target publish-wrangler --base=v3/prod~1 --head=v3/prod
```
jeffywu: Change this to match what is already on v3/dev

packages/shared/web/src/header/use-leverage-earn-links.tsx
matthew-garrett: Are you fine with having the useSelectedNetwork defaulting to mainnet if doesn't exist? Or do you want the same functionality as the useDefaultNetwork back?

matthew-garrett: I am in favor of sticking with mainnet only from the landing page. If we do it based on the useWalletConnectedNetwork hook it will cause a refresh of the app if there is a connected wallet when the app loads.

packages/features/portfolio/src/side-drawers/cool-down-drawer.tsx
```diff
@@ -21,19 +21,17 @@ import {
...
 import { SNOTEWeightedPool } from '@notional-finance/core-entities';
 
 export const CoolDownDrawer = () => {
   const theme = useTheme();
   const account = useAccountDefinition(Network.mainnet);
-  const selectedNetwork = useSelectedNetwork();
   const { isReadOnlyAddress, onSubmit, transactionStatus, transactionHash } =
     useTransactionStatus(Network.mainnet);
   const { days, coolDownEnd, coolDownBegin } = useCoolDownDrawer();
-  const walletConnectedNetwork = useWalletConnectedNetwork();
```
jeffywu: I'm not sure this will work, the selected network will always ben mainnet in this case, we would need to check whether or not their wallet is connected to mainnet.

jeffywu: Maybe we should maintain the useWalletConnectedNetwork but just expose a function that calls onboard to get the current network inside a callback or something.

packages/notionable/src/models/app-store.ts
```diff
@@ -1,24 +1,20 @@
...
   .actions((self) => ({
     fetchKpiData: flow(function* () {
       try {
-        const response = yield fetch(
-          `${
-            process.env['NX_REGISTRY_URL'] ||
-            'https://registry.notional.finance'
-          }/all/kpi`
-        );
+        const response = yield fetch(`${process.env['NX_DATA_URL']}/kpi`);
```
jeffywu: The deleted code is actually correct here.

packages/notionable/src/models/app-store.ts
```diff
@@ -57,4 +59,14 @@ export const AppStore = AppState.create({
     totalDeposits: 0,
     totalOpenDebt: 0,
   },
-});
+  wallet: { 
+    isSanctionedAddress: false,
+    isAccountPending: false,
+    userWallet: {
+      selectedChain: Network.mainnet,
```
jeffywu: We should leave `selectedChain` undefined, or better just remove it if it is not being used.

packages/notionable/src/models/wallet-store.ts
```diff
@@ -0,0 +1,39 @@
+import { types, Instance } from 'mobx-state-tree';
+import { Network } from '@notional-finance/util';
+import { AccountState } from '../global';
+
+const UserWalletModel = types.model('UserWalletModel', {
+  selectedChain: types.optional(types.enumeration('Network', Object.values(Network)), Network.mainnet),
```
jeffywu: The default value should be undefined

packages/shared/notionable-hooks/src/use-transaction.tsx
```diff
@@ -91,7 +103,9 @@ export function useTransactionStatus(network: Network | undefined) {
   const { isReadOnlyAddress, submitTransaction } = useSubmitTransaction();
   const { pathname } = useLocation();
   const isWalletConnectedToNetwork =
-    !!network && !!walletNetwork && network === walletNetwork;
+    !!network &&
+    !!userWallet?.selectedAddress &&
+    network === userWallet?.selectedAddress;
```
jeffywu: This comparison looks wrong

packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -49,34 +50,34 @@ export function useWalletConnected() {
...
 }
 
 export function useWalletConnectedNetwork() {
   const {
-    globalState: { wallet },
-  } = useNotionalContext();
-  return wallet?.selectedChain;
+    wallet: { userWallet },
+  } = useAppStore();
+  return userWallet?.selectedChain;
```
jeffywu: The issue with this is that I'm not sure that onboard will always properly trigger an update inside the appstore for the user wallet. We may want to use some sort of async method to get the current selected chain inside the wallet.

packages/shared/trade/src/transaction-sidebar/transaction-sidebar.tsx
```diff
@@ -87,27 +86,28 @@ export const TransactionSidebar = ({
...
+    wallet: { userWallet },
+  } = useAppStore();
   const isBlocked = useLeverageBlock();
   const approvalData = useTransactionApprovals(
     context,
     requiredApprovalAmount,
     variableBorrowRequired
   );
-  const mustSwitchNetwork = selectedNetwork !== walletConnectedNetwork;
+  const mustSwitchNetwork = selectedNetwork !== userWallet?.selectedChain;
```
jeffywu: Can't you just keep the old hooks? Do you need to refactor it this way?

packages/shared/web/src/header/use-leverage-earn-links.tsx
jeffywu: That's fine as long as the links work

packages/shared/notionable-hooks/src/use-wallet.ts
```diff
@@ -49,34 +50,34 @@ export function useWalletConnected() {
 
 export function useWalletAddress() {
   const {
-    globalState: { wallet },
-  } = useNotionalContext();
+    wallet: { userWallet },
+  } = useAppStore();
 
-  return wallet?.selectedAddress;
+  return userWallet?.selectedAddress;
```
jeffywu, line 57: Should these be wrapped in useObserver? or is that already done in useAppStore?

packages/shared/trade/src/transaction-sidebar/transaction-sidebar.tsx
matthew-garrett: We're using the selectedChain and isReadOnlyAddress so I thought I'd just use the userWallet instead of having two hooks

packages/shared/notionable-hooks/src/use-wallet.ts
matthew-garrett: Not exactly sure what you're wanting here. We could always just use the onboard hooks here and return the chain directly

matthew-garrett: The useEffect in the use-connect hook should keep the userWallet up to date how it has in the past though

packages/features/portfolio/src/side-drawers/cool-down-drawer.tsx
matthew-garrett: I brought back the old hook that I updated to use the new context

packages/shared/notionable-hooks/src/use-wallet.ts
matthew-garrett: I am hesitant to throw useObserver around in the hooks. I am in favor of using observer at the component level. I think it makes it more obvious what is supposed to be rendering and what is not. It'd be easy to forget which hooks use useObserver and which do not.

packages/shared/trade/src/transaction-sidebar/transaction-sidebar.tsx
matthew-garrett: If you want I can bring the old hooks back

packages/shared/notionable-hooks/src/use-wallet.ts
jeffywu: yeah use the onboard hooks instead

packages/shared/trade/src/transaction-sidebar/transaction-sidebar.tsx
jeffywu: lets use the old hooks

packages/shared/notionable-hooks/src/use-wallet.ts
jeffywu: ok thats fine

PR #1081, opened 2024-08-31
Bugfix/grab bag

matthew-garrett:
- Fixes issue of user not being able to see portfolio
- Fixes earnings at maturity calculation on portfolio holdings
- Refactors vault reinvestment dates
- Fixes index error in side nav

### DIFF
--- a/packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
+++ b/packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
@@ -33,6 +33,8 @@ import {
   PRIME_CASH_VAULT_MATURITY,
   pointsMultiple,
   Network,
+  getDateString,
+  SECONDS_IN_DAY,
 } from '@notional-finance/util';
 import { VaultAccountRiskProfile } from '@notional-finance/risk-engine';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -85,12 +86,12 @@ export function getVaultReinvestmentDates(
       ? reinvestmentData[vaultAddress]
       : undefined;
   if (vaultReinvestmentData && vaultReinvestmentData.length > 0) {
-    arbReinvestmentDate = moment(vaultReinvestmentData[0].timestamp * 1000)
-      .add(1, 'days')
-      .format('MMM DD YYYY');
-    mainnetReinvestmentDate = moment(vaultReinvestmentData[0].timestamp * 1000)
-      .add(7, 'days')
-      .format('MMM DD YYYY');
+    arbReinvestmentDate = getDateString(vaultReinvestmentData[0].timestamp);
```
jeffywu: This is wrong. We add 1 day to the arbReinvestment date above.

packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -85,12 +86,12 @@ export function getVaultReinvestmentDates(
...
+    mainnetReinvestmentDate = getDateString(
```
jeffywu, line 92: Why are the two getDateString calls different?

matthew-garrett: I pushed a fix this for this. Lmk if it's what you're expecting

PR #1083, opened 2024-09-02
Feat/add yield data

jeffywu:


packages/shared/web/src/dashboard-view/hooks/use-variable-rate-grid.tsx
```diff
@@ -20,36 +23,45 @@ export const useVariableRateGrid = (
   network: Network | undefined,
   product: PRODUCTS
 ) => {
-  const {
-    yields: { variableLend, variableBorrow },
-  } = useAllMarkets(network);
+  // TODO: this can be put into a hook?
```
jeffywu, line 26: @matthew-garrett Can you take a look at this and see how you want to integrate this into react? Once you work out a pattern we can use this to replace all the dashboard views.

PR #1095, opened 2024-09-04
Vault apy claim simulation updates

T-Woodward:
Updates the incentive claim simulation methodology to work around observed issues for Aura, ConvexMainnet, and ConvexArbitrum pools. For Aura and ConvexMainnet pools, before warping forward we check to see where the incentive accrual period finish point is and make sure we're at least one day behind it. For ConvexArbitrum pools, we make sure that we're either at least 2 hours ahead of the most recent Thursday at midnight UTC or we're at least one day behind the nearest upcoming Thursday midnight UTC.

### DIFF
--- a/apps/vault-apy/src/APYSimulator.ts
+++ b/apps/vault-apy/src/APYSimulator.ts
@@ -84,47 +84,13 @@ export default class APYSimulator {
     for (let i = 1; i <= numOfDays; i++) {
       const currentTimestamp = startingTimestamp - i * ONE_DAY_IN_SECONDS;
       log(`processing day ${i}, ${currentTimestamp}`);
-      const forkBlock = await this.#getBlockAtTimestamp(currentTimestamp);
+      const initialForkBlock = await this.#getBlockAtTimestamp(currentTimestamp);
 
       for (const vault of this.#config.vaults) {
         log(`Processing vault: ${vault.address}`);
         try {
-          if (vault.rewardPoolType === RewardPoolType.Aura || vault.rewardPoolType === RewardPoolType.ConvexMainnet) {
-            const periodFinish = await this.getPeriodFinishForVault(vault.address, forkBlock);
-            const forkBlockTimestamp = (await this.#alchemyProvider.getBlock(forkBlock)).timestamp;
-            const timeDifference = periodFinish - forkBlockTimestamp;
-            log(`Time difference between periodFinish and forkBlock: ${timeDifference} seconds`);
-
-            if (timeDifference < -ONE_DAY_IN_SECONDS || timeDifference > ONE_DAY_IN_SECONDS) {
-              // Continue as normal
-            } else {
-              // Find the fork block 1 day behind periodFinish
-              const adjustedForkBlock = await this.#getBlockAtTimestamp(periodFinish - ONE_DAY_IN_SECONDS);
-              log(`Adjusted forkBlock to ${adjustedForkBlock} (1 day behind periodFinish)`);
-              await this.run(adjustedForkBlock, vault.address);
-              continue; // Skip the normal run call below
-            }
-          } else if (vault.rewardPoolType === RewardPoolType.ConvexArbitrum) {
-            const forkBlockTimestamp = (await this.#alchemyProvider.getBlock(forkBlock)).timestamp;
-            const mostRecentThursdayMidnight = this.#getMostRecentThursdayMidnight(forkBlockTimestamp);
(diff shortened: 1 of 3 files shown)


apps/vault-apy/src/APYSimulator.ts
```diff
@@ -80,13 +80,55 @@ export default class APYSimulator {
...
     for (let i = 1; i <= numOfDays; i++) {
-      log(`processing day ${i}, ${startingTimestamp - i * ONE_DAY_IN_SECONDS}`);
-      const forkBlock = await this.#getBlockAtTimestamp(
-        startingTimestamp - i * ONE_DAY_IN_SECONDS
-      );
-
-      await this.run(forkBlock);
+      const currentTimestamp = startingTimestamp - i * ONE_DAY_IN_SECONDS;
+      log(`processing day ${i}, ${currentTimestamp}`);
+      const forkBlock = await this.#getBlockAtTimestamp(currentTimestamp);
```
jeffywu: It looks like most of this logic is about adjusting the fork block before calling `this.run`. I think this code would be cleaner if you created a function called `getForkBlock` and then put all your adjustments in there. Then the code inside the loop would look like:

```
for (const vault of ... ) {
   const forkBlock = getForkBlock(vault, currentTimestamp)
   await this.run(forkBlock, vault.address)
}
```

apps/vault-apy/src/APYSimulator.ts
```diff
@@ -95,75 +137,125 @@ export default class APYSimulator {
...
+    }
+
     for (let i = 0; i <= numOfDays; i++) {
       log(`processing day ${i}`);
-      const forkBlock = await this.#getBlockAtTimestamp(
+      let forkBlock = await this.#getBlockAtTimestamp(
         startingTimestamp - i * ONE_DAY_IN_SECONDS
       );
 
+      if (vaultData.rewardPoolType === RewardPoolType.Aura || vaultData.rewardPoolType === RewardPoolType.ConvexMainnet) {
```
jeffywu: Without doing a diff, it looks like these adjustments are the same as above, I would put it into a function and re-use the code. If it's different, it's still worth breaking out into a function.

apps/vault-apy/src/APYSimulator.ts
```diff
@@ -563,4 +655,101 @@ export default class APYSimulator {
...
+    
+    if (!vaultData) {
+      throw new Error(`Vault address ${vaultAddress} not found in config`);
+    }
+
+    if (!vaultData.gauge) {
+      throw new Error(`Gauge address not specified for vault ${vaultAddress}`);
+    }
+
+    const provider = new ethers.providers.JsonRpcProvider(this.#config.alchemyUrl);
```
jeffywu: Can't you just refer to `this.#alchemyProvider` here instead?

apps/vault-apy/src/APYSimulator.ts
```diff
@@ -563,4 +655,101 @@ export default class APYSimulator {
...
+
+    // ABI fragment for the periodFinish variable
+    const gaugeABI = [
+      "function periodFinish() view returns (uint256)"
```
jeffywu: There's a file called `interfaces.ts` where this declaration should go.

apps/vault-apy/src/APYSimulator.ts
```diff
@@ -563,4 +655,101 @@ export default class APYSimulator {
...
+    ];
+
+    const gaugeContract = new Contract(vaultData.gauge, gaugeABI, provider);
+
+    try {
+      // Call the periodFinish function at the specified block
+      const periodFinish: BigNumber = await gaugeContract.callStatic.periodFinish({
```
jeffywu: You don't need the `callStatic` modifier here.

apps/vault-apy/src/APYSimulator.ts
```diff
@@ -563,4 +655,101 @@ export default class APYSimulator {
...
+      return;
+    }
+
+    const latestBlock = await this.#alchemyProvider.getBlockNumber();
+    log(`Processing all vaults at block ${latestBlock}`);
+
+    for (const vault of this.#config.vaults) {
+      log(`Processing vault: ${vault.address}`);
+      try {
+        let forkBlock = latestBlock;
```
jeffywu: Make your adjustments inside functions as much as possible, it will make maintaining this code a lot easier. If there are a slight differences in these adjustments with the others, try to fit it into logic groupings so that future adjustments can be done easily without much chance of breaking other functionality.

PR #1096, opened 2024-09-04
Feat/lev liquidity dash mobx

matthew-garrett:


packages/shared/web/src/dashboard-view/hooks/use-network-tokens.tsx
```diff
@@ -3,15 +3,51 @@ import { Network } from '@notional-finance/util';
...
-  const yieldData = model.getTokensByType(tokenType).map((t) => ({
-    token: t,
-    apy: model.getSpotAPY(t.id),
-    tvl: model.getTVL(t),
-    liquidity: model.getLiquidity(t),
-    underlying: t.underlying ? model.getTokenByID(t.underlying) : undefined,
-  }));
+  let yieldData: any[] = [];
+
+  if (options?.leveragedNToken) {
```
jeffywu: I don't think we should use if/else statements here. Can we just make different hooks instead? It will be a lot cleaner and less likely to have bugs

jeffywu: Also by declaring this as `any` you lose all the type checking downstream.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-list.tsx
```diff
@@ -189,78 +180,70 @@ export const useLiquidityList = (
       const walletBalance = account
         ? account.balances.find((t) => t.tokenId === y.underlying.id)
         : undefined;
-      const organicApy = (y.organicAPY || 0) + (y.feeAPY || 0);
+      const organicApy = (y.apy.organicAPY || 0) + (y.apy.feeAPY || 0);
```
jeffywu, line 183: Let's change the YieldViews.ts file to do this math upstream and remove it here. I'd like to remove as much logic like this as posssible from the components.

PR #1101, opened 2024-09-09
fix: wrong timestamp saved to db

T-Woodward:


apps/vault-apy/src/APYSimulator.ts
```diff
@@ -288,12 +294,12 @@ export default class APYSimulator {
...
       network: this.#network,
-      date: new Date(prevBlock.timestamp * 1000).toISOString(),
+      date: new Date(originalTimestamp * 1000).toISOString(),
       ///////////////////////////////////////////////////////////////////////
 
       swapFees: poolFeesInPrimary.toString(),
-      blockNumber: prevBlock.number,
-      timestamp: floorToMidnight(prevBlock.timestamp),
+      blockNumber: forkBlock,
+      timestamp: floorToMidnight(originalTimestamp),
```
T-Woodward, line 302: @jeffywu this would save the unadjusted timestamp but the adjusted block, so they wouldn't be in sync in the event that you had to adjust the forkBlock. Could use the original block as well, but not sure if that's better

PR #1107, opened 2024-09-12
Feat/portfolio state zero mobx

jeffywu:


packages/features/portfolio/src/containers/portfolio-state-zero/hooks/use-card-data.tsx
```diff
@@ -9,73 +8,27 @@ import {
...
+import { sumAndFormatIncentives } from '@notional-finance/shared-web/dashboard-view/hooks/utils';
+import { getAvailableVaults } from './use-network-token-data';
 
 export const useCardData = (
-  selectedTab: number,
+  selectedTabIndex: number,
   activeToken: string,
-  allTokenData: ReturnType<typeof useTokenData>['allTokenData'],
-  availableSymbols: string[]
+  tokenData: any,
```
jeffywu: Using `any` here is not good, you should be able to get the return type using typescript utility functions. If not, use the AI helper to generate the proper types.

packages/features/portfolio/src/containers/portfolio-state-zero/state-zero-data.tsx
```diff
@@ -0,0 +1,184 @@
...
+  active: boolean;
+}
+
+export const StateZeroData = ({
+  productGroupData,
+  defaultSymbol,
+  tokenList,
+  selectedTabIndex,
+}: {
+  productGroupData: any[];
```
jeffywu: This needs to be properly typed.

packages/notionable/src/hooks/use-portfolio-store.tsx
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { PortfolioStoreType } from '../stores/portfolio-store';
+import { RootStoreContext } from '../stores/root-store';
+
+export const usePortfolioStore = (): PortfolioStoreType => {
```
jeffywu: If these are all coming off the root store, I think you should just put it all in the same file, `use-root-store`

packages/notionable/src/stores/root-store.ts
```diff
@@ -0,0 +1,82 @@
...
+        return self.mainnetStore
+      case Network.arbitrum:
+        return self.arbitrumStore
+      default:
+        return self.allNetworksStore
+    }
+  }
+  return { getNetworkClient }
+})
+  export const createRootStore = (): RootStoreType => {
```
jeffywu: Indentation here is weird.

packages/notionable/src/stores/root-store.ts
```diff
@@ -0,0 +1,82 @@
...
+const userSettings = getFromLocalStorage('userSettings');
+
+const RootStore = types.model("RootStore", {
+  mainnetStore: NetworkClientModel, 
+  arbitrumStore: NetworkClientModel, 
+  allNetworksStore: NetworkClientModel, 
+  portfolioStore: PortfolioStoreModel,
+  appStore: AppStoreModel,
+}).actions(self => {
+  const getNetworkClient = (network: Network) => {
```
jeffywu: Why is this an action? Shouldn't it be a view?

packages/notionable/src/stores/root-store.ts
```diff
@@ -0,0 +1,82 @@
...
+    switch (network) {
+      case Network.mainnet:
+        return self.mainnetStore
+      case Network.arbitrum:
+        return self.arbitrumStore
+      default:
```
jeffywu: All networks is not default, is is Network.all, throw an error on default like the other function

packages/notionable/src/stores/root-store.ts
```diff
@@ -0,0 +1,82 @@
...
+      mainnetStore: {
+        network: Network.mainnet
+      },
+      arbitrumStore: {
+        network: Network.arbitrum
+      },
+      allNetworksStore: {
+        network: Network.all
+      },
+      portfolioStore: {
```
jeffywu: When you define these stores, it's better to embed the default vault into the model definition using type.optional. Some of these values should also be undefined to signify that they haven't loaded, like `heroStats` for example.

packages/shared/notionable-hooks/src/context/AppContext.ts
```diff
@@ -1,12 +1,12 @@
-import { AppStoreType } from '@notional-finance/notionable';
+import { AppStoreType, RootStoreContext } from '@notional-finance/notionable';
 import { useContext, createContext } from 'react';
 
 export const AppContext = createContext<AppStoreType | null>(null);
 
 export const useAppStore = (): AppStoreType => {
-  const context = useContext(AppContext);
```
jeffywu, line 7: Move this to the use-root-store file.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-list.tsx
```diff
@@ -178,16 +178,16 @@ export const useLiquidityList = (
   const listData = yieldData
     .map((y) => {
       const walletBalance = account
-        ? account.balances.find((t) => t.tokenId === y.underlying.id)
```
jeffywu, line 181: It seems to be like there is something wrong with the types here. If `y` is undefined then the entire return struct is basically empty

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
     }
   };
 
+  const getUniqueUnderlyingSymbols = (productGroupData: any[][]) => {
```
jeffywu: It's very bad practice to type things as `any`, especially inside library functions like this. Use the AI helpers to generate proper types.

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+    const uniqueUnderlyingSymbols = productGroupData
+      .flat()
```
jeffywu, line 450: There is already a `unique` function in `array-helpers`, we should use that instead of creating a special one off function here.

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+      productGroupData: productGroupData,
+      defaultSymbol,
+    };
+  };
+
+  const getPortfolioStateZeroEarnData = () => {
+    const group = ['fCash', 'PrimeCash', 'nToken'];
+    const productGroupData = group.map((group) => {
+      return getTokensByType(group)
+        .filter((data) => !data?.isFCashDebt)
```
jeffywu, line 515: Isn't this filter already applied inside getTokensByType?

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+      }
+
+      return highestAPYSymbol;
+  };
+
+  const getPortfolioStateZeroBorrowData = () => {
+    const group = ['fCash', 'PrimeDebt'];
+    const productGroupData = group.map((group) => {
+      return getTokensByType(group)
+        .filter((data) => data?.isFCashDebt || data?.tokenType === 'PrimeDebt')
```
jeffywu, line 488: Unclear why this filter is necessary? Shouldn't this already be done in getTokensByType

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+    return {
+      tokenList,
+      productGroupData,
+      defaultSymbol: highestAPYSymbol,
+    };
+  };
+
+  const getLeveragedNTokenData = () => {
+    const leveragedNTokenData = getTokensByType('nToken').map((t) => {
+          const debtTokens = getDefaultLeveragedNTokenAPYs(t);
```
jeffywu: indentation here is wrong, you should get prettier installed in vscode and working. Otherwise if I touch a file it reformats the whole thing and it looks like there were a ton of changes.

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+          const leveragedNTokenData = debtTokens.reduce((max, current) => {
+            return current?.apy?.totalAPY &&
+              max?.apy?.totalAPY &&
+              current.apy.totalAPY > max.apy.totalAPY
+              ? current
+              : max;
+          }, debtTokens[0]);
```
jeffywu: This will throw an error if debTokens has a length of 0. You can use the library function `firstValue` here instead.

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+          underlying: t.underlying ? getTokenByID(t.underlying) : undefined,
+          collateralFactor: getDebtOrCollateralFactor(t, false),
+          debtToken: leveragedNTokenData?.debtToken
+        };
+      });
+      return leveragedNTokenData
+  }
+
+  const getAllListedVaultsData = () => {
+    const listedVaults = getAllListedVaults(true)
```
jeffywu: Why pass in the boolean here? It's already defaulted to true.

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -430,6 +444,188 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+    const allListedVaultsData = getTokensByType('VaultShare')
+    .filter((t) => t.vaultAddress && listedVaultAddresses.includes(t.vaultAddress))
+    .map((t) => {
+          const data = getDefaultVaultAPYs(t.vaultAddress || '');
+          const vaultAPYs = data.reduce((max, current) => {
+            return (current.apy.totalAPY || 0) > (max.apy.totalAPY || 0)
+              ? current
+              : max;
+          }, data[0])?.apy;
+          const vaultDebtToken = data.reduce((max, current) => {
```
jeffywu: Why do the reduce twice? Isn't the correct debtToken already sitting in the vaultAPYs data structure?

packages/shared/web/src/dashboard-view/hooks/use-network-tokens.tsx
```diff
@@ -26,10 +27,10 @@ export const useNetworkTokens = (
   return yieldData;
 };
 
-export const useLeveragedNTokens = (network: Network | undefined) => {
-  const model = getNetworkModel(network);
-  const yieldData = model.getTokensByType('nToken').map((t) => {
-    const debtTokens = model.getDefaultLeveragedNTokenAPYs(t);
+export const useLeveragedNTokens = () => {
```
jeffywu, line 31: It seems to me like the state zero code and the code to get the data for the dashboard views are heavily duplicated. I think it would be best to consolidate the logic in both places into a single code path.

packages/shared/web/src/dashboard-view/hooks/use-liquidity-list.tsx
matthew-garrett: Will address in my next PR

PR #1114, opened 2024-09-14
Feat/vault dashboard mobx

matthew-garrett:
- Adds Vaults data to dashboard pages
- Breaks up YieldView functions to be able to be used by portfolio state zero and dashboard pages


packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -570,9 +599,13 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
 
     const allListedVaultsData = getTokensByType('VaultShare')
     .filter((t) => t.vaultAddress && listedVaultAddresses.includes(t.vaultAddress))
+    .filter((t, index, self) =>
```
jeffywu, line 602: What is the point of doing this filter?

packages/core-entities/src/models/views/YieldViews.ts
```diff
@@ -603,6 +637,29 @@ export const YieldViews = (self: Instance<typeof NetworkModel>) => {
...
+
+    return {
+      tokenList,
+      productGroupData,
+      defaultSymbol: highestAPYSymbol,
+    };
+  };
+
+  const getPortfolioStateZeroBorrowData = () => {
+    const productGroupData = [getPrimeCashDebtData(), getFCashDebtData()];
```
jeffywu, line 653: Instead of making this an array of arrays, why not use a key value structure?

packages/core-entities/src/models/views/YieldViews.ts
matthew-garrett: This filter was to remove duplicates of vaultAddresses

packages/core-entities/src/models/views/YieldViews.ts
matthew-garrett: These are are moved to the portfolio store in the next PR. They are in a array because there are times that we need to loop over all of them and add them up or do different data manipulation and the array made it easier

PR #1131, opened 2024-09-19
build: adds routes to root store

matthew-garrett:
- Adds route to rootStore
- Adds ROUTE_MATCH enum
- Adds useEffect to set the current route in AppLayoutRoutes
- Adds walletStore to rootStore and removes it from appStore
- Adds conditional to portfolioStore based on route in rootStore

### DIFF
--- a/apps/web/src/containers/TradingContest/hooks/use-leaderboard-data.ts
+++ b/apps/web/src/containers/TradingContest/hooks/use-leaderboard-data.ts
@@ -3,7 +3,7 @@ import {
   formatNumberAsPercent,
   truncateAddress,
 } from '@notional-finance/helpers';
-import { useAppStore } from '@notional-finance/notionable';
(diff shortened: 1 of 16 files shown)


apps/web/src/containers/TradingContest/hooks/use-leaderboard-data.ts
```diff
@@ -94,7 +94,7 @@ export function useLeaderboardData() {
   const [highRollerPartner, setHighRollerPartner] = useState<number>(0);
   const [fatCatPartner, setFatCatPartner] = useState<number>(0);
   const network = useSelectedNetwork();
-  const {wallet: {userWallet}} = useAppStore(); 
+  const { walletStore: { userWallet } } = useRootStore();
```
jeffywu: Can we make a hook for the walletStore as well?

PR #1140, opened 2024-09-23
feat: check percentage when trading before reinvestment

sbuljac:
Before processing with quote received from ox API check that price is withing acceptable percentage from Defillama price

### DIFF
--- a/apps/rewards/src/checkPercentageLoss.ts
+++ b/apps/rewards/src/checkPercentageLoss.ts
@@ -1,0 +1,71 @@
+import { Network } from '@notional-finance/util';
+import { Env } from './types';
+import { BigNumber } from 'ethers';
+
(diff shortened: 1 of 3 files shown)


sbuljac:
@T-Woodward  By "Check this loss against the acceptable percentage loss table. If the calculated loss is greater than the loss value from the table, skip this run and try again in 2 hours." do you mean to skip reinvestment only for reward token that has large loss or completely skip reinvestment for all rewards tokens that we are trying to sell/reinvest

Ref: https://docs.google.com/document/d/1OwoxvoHbL4QIYMa9gEAav4oLzg7MwXU7BFGN_EMnJv4/edit?pli=1

T-Woodward:
@sbuljac I think it's cleaner to skip the entire reinvestment for a specific vault if any one of the trades fails this price check

PR #1141, opened 2024-09-23
refactor: move token addresses to util folder

sbuljac:


packages/core-entities/src/client/configuration-client.ts
```diff
@@ -120,6 +120,9 @@ export class ConfigurationClient extends ClientRegistry<AllConfigurationQuery> {
   getAllListedVaults(network: Network, includeDisabled = false) {
     return this.getLatestFromSubject(network, network)
       ?.vaultConfigurations.filter((v) => v.enabled || includeDisabled)
+      .filter(
```
jeffywu: @sbuljac is there a reason why you added this?

packages/util/src/constants.ts
```diff
@@ -141,3 +152,72 @@ export const ACCOUNT_ID_RANGES = [
...
+    USDCe: '0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8',
+    USDC: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831',
+    USDT: '0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9',
+    wstETH: '0x5979D7b546E38E414F7E9822514be443A4800529',
+    WBTC: '0x2f2a2543B76A4166549F7aaB2e75Bef0aefC5B0f',
+    FRAX: '0x17FC002b466eEc40DaE837Fc4bE5c67993ddBd6F',
+  },
+};
+
+export const vaults = {
```
jeffywu, line 198: We also have this function in core-entities called `whitelistedVaults` we should probably combine with this data structure somehow

PR #1154, opened 2024-09-28
feat: add new Tenderly action for process loss on reinvestment trades

sbuljac:
Already deployed and tested


package.json
```diff
@@ -50,7 +50,7 @@
     "dotenv": "^16.0.2",
     "echarts": "^5.5.0",
     "echarts-for-react": "^3.0.2",
-    "ethers": "^5.7.1",
+    "ethers": "^5",
```
jeffywu: Do we need to change this dependency globally? Just wondering if this will break something elsewhere?

sbuljac: no need, I will change it to the old value

PR #1160, opened 2024-10-01
Feat/use market mobx

matthew-garrett:
- Refactors use-market file hooks
- Deletes leveraged lending UI


apps/web/src/containers/TradingContest/points-dashboard/hooks/use-qualifying-products-table-dropdowns.tsx
```diff
@@ -9,14 +9,13 @@ import {
...
 } from '@notional-finance/icons';
-import { Network } from '@notional-finance/util';
 import { TokenIcon } from '@notional-finance/icons';
 import { useAllUniqueUnderlyingTokens } from '@notional-finance/notionable-hooks';
 import { useTheme } from '@mui/material';
 
 export const useQualifyingProductsTableDropdowns = () => {
   const theme = useTheme();
-  const depositTokens = useAllUniqueUnderlyingTokens([Network.arbitrum]);
+  const depositTokens = useAllUniqueUnderlyingTokens();
```
jeffywu: This hook looks wrong, what would be the use case for looking across all networks?

packages/features/borrow/src/borrow-fixed/hooks/use-borrow-fixed-multi-chart.tsx
```diff
@@ -18,7 +18,7 @@ export const useBorrowFixedMultiChart = () => {
     true
   );
   const { selectedfCashId, onSelect } = useMaturitySelect('Debt', context);
-  const nToken = useNToken(deposit?.network, deposit?.currencyId);
+  const nToken = useNToken(deposit?.currencyId);
```
jeffywu: We should try to just refactor this hook out of the codebase altogether, maybe look at changing the useChart hook if we're passing it in here.

packages/features/lend/src/lend-fixed/hooks/use-totals-data.tsx
```diff
@@ -5,23 +5,20 @@ import {
...
 export const useTotalsData = (
   deposit: TokenDefinition | undefined,
   collateral: TokenDefinition | undefined,
   baseCurrency: FiatKeys
 ) => {
-  const {
-    yields: { fCashLend },
-  } = useAllMarkets(deposit?.network);
+  const currentNetworkStore = useCurrentNetworkStore();
+  const fCashLend = currentNetworkStore.getAllFCashYields();
```
jeffywu: Inside of these use-totals-data hooks instead of fetching all of the yields and then filtering, we should call a function that gets exactly the values we need. If you look lower down the component it looks like we are filtering out for the TVL, we can get that value directly from the store instead of doing things the old way.

apps/web/src/containers/TradingContest/points-dashboard/hooks/use-qualifying-products-table-dropdowns.tsx
matthew-garrett: This doesn't matter any more because we are not using this table.

PR #1186, opened 2024-10-07
build: adds pendle how it works

matthew-garrett:


packages/util/src/enums.ts
```diff
@@ -26,6 +26,7 @@ export enum PRODUCTS {
 export enum REINVESTMENT_TYPE {
```
jeffywu: PendlePT is not a reinvestment type, instead of creating a new enum here I think you should just use the `type VaultType` from the core-entities repo instead.

matthew-garrett: My bad here

matthew-garrett: If we use VaultType we'll have to do string comparisons for the conditionals. Are you fine with that?

PR #1196, opened 2024-10-08
build: adds DeBankNetWorth to intercom

matthew-garrett:


packages/shared/wallet/.eslintrc.json
```diff
@@ -18,7 +18,9 @@
     },
     {
       "files": ["*.ts", "*.tsx"],
-      "rules": {}
+      "rules": {
```
jeffywu: Instead of turning this off, you should add the words to the word list by installing the cspell extension

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -1,16 +1,63 @@
...
+    !userSettings.debankAddress &&
+    selectedAccount &&
+    !isReadOnlyAddress &&
+    netWorth === 0
+  ) {
+    fetchDeBankNetWorth(selectedAccount).then((netWorth) => {
+      setNetWorth(Math.trunc(netWorth));
+      setInLocalStorage('userSettings', {
+        ...userSettings,
+        debankAddress: selectedAccount,
```
jeffywu: Why do we need to set either of these in local storage? The net worth will change over time.

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -1,16 +1,63 @@
...
   const { update } = useIntercom();
+  const userSettings = getFromLocalStorage('userSettings');
+  const [netWorth, setNetWorth] = useState<number>(0);
   const {
     globalState: { networkAccounts, wallet },
   } = useNotionalContext();
   const isReadOnlyAddress = wallet?.isReadOnlyAddress;
   const selectedAccount = wallet?.selectedAddress;
 
+  if (
```
jeffywu: This should also be inside a useEffect, I don't think its good to run promises inside the render cycle like this

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -1,16 +1,63 @@
...
+import {
+  getFromLocalStorage,
+  setInLocalStorage,
+  SupportedNetworks,
+} from '@notional-finance/util';
 import { useIntercom } from 'react-use-intercom';
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
+
+const DeBankAPIKey = process.env['NX_DE_BANK_API_KEY'] as string | undefined;
```
jeffywu: it should be "NX_DEBANK_API_KEY", debank is one word

matthew-garrett: Typescript freaked out about this being DEBANK. I had it that originally. When I separated the words it was fine

packages/shared/wallet/src/hooks/use-intercom-update.tsx
matthew-garrett: The debankAddress is in local storage to prevent excess api calls because we have to pay for them. The debankNetworth is in local storage so that if someone disconnects their wallet and connects again and the connection is recorded in intercom the debankNetworth is still sent and not zero

packages/shared/wallet/src/hooks/use-intercom-update.tsx
matthew-garrett: I kept this out of the useEffect because I think it will likely cause more renders being inside than in. I'll go ahead and move it in if you want

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -1,22 +1,101 @@
...
+
+  const [isFetching, setIsFetching] = useState(false);
+
+  const debouncedFetchDeBankNetWorth = useCallback(
+    async (address: string) => {
+      if (isFetching) return;
+      setIsFetching(true);
+      try {
+        const netWorth = await fetchDeBankNetWorth(address);
+        const currentTimestamp = new Date().toISOString();
```
jeffywu: Use `getNowSeconds() instead.

packages/shared/wallet/src/hooks/use-intercom-update.tsx
```diff
@@ -1,22 +1,101 @@
...
+
+  const updateDeBankWallet =
+    userSettings.debankAddress &&
+    !isReadOnlyAddress &&
+    userSettings.debankAddress !== selectedAccount;
+
+  const weeklyCheck = (() => {
+    if (userSettings.debankTimestamp) {
+      const lastUpdateDate = new Date(userSettings.debankTimestamp);
+      const currentDate = new Date();
```
jeffywu: Use `getNowSeconds()` here as well

packages/shared/wallet/.cspell.json
```diff
@@ -0,0 +1,3 @@
+{
+  "words": ["debank", "leaderboard", "Avenir"]
```
jeffywu: add these words as lines to the notional-words.txt

PR #1212, opened 2024-10-15
build: moves logic to on wallet connect adds created_at to intercom u…

matthew-garrett:
- Adds created_at metric to intercom that will track when a user first connects their wallet. The metric is displayed as signed up in intercom
- Adds data dog and segment wallet connect tracking


packages/shared/wallet/src/wallet-selector/wallet-selector.tsx
```diff
@@ -40,7 +40,7 @@ export function WalletSelector() {
   const [showAlert, setShowAlert] = useState<boolean>(false);
   const { setWalletSideDrawer, clearWalletSideDrawer } = useSideDrawerManager();
   const { openDrawer } = useWalletSideDrawer();
-  useIntercomUpdate();
+  // useWalletTracking();
```
jeffywu: Should this be commented out?

PR #1238, opened 2024-10-28
build: adds debank lender data

matthew-garrett:


packages/notionable/src/global/sagas/on-wallet-connect.ts
```diff
@@ -89,27 +89,68 @@ function onWalletChange$(global$: Observable<GlobalState>) {
...
 
-    return data.total_usd_value;
+    const [balanceData, protocolsData] = await Promise.all([
+      balanceResponse.json(),
+      protocolsResponse.json(),
+    ]);
+
+    let isLender;
+
+    protocolsData?.map((data) => {
```
jeffywu: Instead of `map` use `find` here and change the variable name to `isBorrower`. that makes more sense given this logic. 

the returned value can be `isLender: !isBorrower`

PR #1243, opened 2024-10-30
build: new user boost wip

matthew-garrett:
build: adds products table

build: new rocket img

build: adds boost values to tx and card pages


apps/web/src/containers/NewUserView/index.tsx
```diff
@@ -0,0 +1,365 @@
...
+import discordIcon from '@notional-finance/assets/images/logos/discord-two-tone.svg';
+import twitterIcon from '@notional-finance/assets/images/logos/twitter-two-tone.svg';
+import { colors, useNotionalTheme } from '@notional-finance/styles';
+import { useProductsTable } from './use-products-table';
+import { Network, THEME_VARIANTS } from '@notional-finance/util';
+import { useNetworkToggle } from '../AnalyticsViews/hooks';
+import newcomer from './newcomer.png';
+
+export const NewUserView = () => {
+  const landingTheme = useNotionalTheme(THEME_VARIANTS.LIGHT, 'landing');
```
jeffywu, line 28: Should we redirect them to error if they are not on the whitelist?

packages/common/icons/src/lib/rocket-icon/rocket-icon.tsx
```diff
@@ -0,0 +1,35 @@
...
+export interface RocketIcon extends SvgIconProps {
+  fill?: string;
+}
+
+export function RocketIcon(props: RocketIcon) {
+  return (
+    <SvgIcon viewBox="0 0 20 20" {...props}>
+      <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 146 146">
+        <path
+          fill-rule="evenodd"
```
jeffywu: Can you fix all the fill-rule and clip-rule errors that show up in the console?

packages/features/portfolio/src/containers/portfolio-holdings/use-detailed-holdings.tsx
```diff
@@ -106,9 +117,10 @@ export function useDetailedHoldingsTable() {
           const isDebt = b.isNegative();
           const { icon, formattedTitle, titleWithMaturity, title } =
             formatTokenType(b.token, isDebt);
-          // const pointsPerDay = getPointsPerDay(b);
-          // const totalPoints =
-          //   arbPoints?.find(({ token }) => token === b.tokenId)?.points || 0;
+          const isBoostedToken = checkBoostedToken(
```
jeffywu: Can we call this `isStarterBoost` instead and `checkStarterBoostToken` or something like that

packages/notionable/src/global/account/communities.ts
```diff
@@ -97,3 +97,7 @@ export function checkSanctionedAddress(account: string) {
   );
   return sanctionList['isSanctioned'](account) as Promise<boolean>;
 }
+
+export function checkNewUserAddress(account: string) {
+  return newUserAddresses.includes(account);
```
jeffywu: Make sure that all the account address is lower case here and all the addresses in the json are lower case as well.

packages/notionable/src/global/global-state.ts
```diff
@@ -117,6 +117,8 @@ interface AddressState {
   communityMembership?: Community[];
   /** Checks if the address is sanctioned on wallet connection */
   isSanctionedAddress: boolean;
+  /** Checks if the address is a new user on wallet connection */
+  isNewUser: boolean;
```
jeffywu: Can we change this to `isStarterBoostUser` instead?

packages/util/src/constants.ts
```diff
@@ -48,6 +49,10 @@ export enum Network {
   arbitrum = 'arbitrum',
 }
 
+export const boostEndDate = new Date('2024-12-16');
```
jeffywu, line 52: I would only export one of these two variables depending on which one you are using.

packages/util/src/constants.ts
```diff
@@ -261,3 +266,3274 @@ export const WETHAddress: Record<Network, Lowercase<TokenAddress>> = {
   [Network.mainnet]: toLower(tokens.mainnet.WETH),
   [Network.arbitrum]: toLower(tokens.arbitrum.WETH),
 };
+
+export const NEW_USER_ADDRESSES = [
```
jeffywu: I think we should move this into a file that we fetch via HTTP instead of committing to the repo

jeffywu: @matthew-garrett can you fetch the list from here:
https://registry.notional.finance/newcomer-boost.json

PR #1256, opened 2024-11-06
Perf/portfolio fixes

matthew-garrett:
- Fixes portfolio page loading issues
- Gets portfolio holdings and txn history pages working

### DIFF
--- a/packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
+++ b/packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import {
   useAccountHistoryChart,
   useSelectedNetwork,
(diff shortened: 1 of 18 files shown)


packages/features/portfolio/src/containers/portfolio-overview/portfolio-overview.tsx
```diff
@@ -92,7 +92,7 @@ const PortfolioOverview = () => {
               }
             />
           )}
-          {vaultHoldingsData.length > 0 && (
+          {vaultHoldingsData[0]?.vault.symbol && (
```
jeffywu: Is the array always guaranteed to have an entry? Is it the "total row"? Maybe we should make a view method on the PortfolioModel that returns a boolean if the account has vault holdings. Won't this same issue happen if the account does not have any portfolio holdings? Maybe we should do the same there.

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -110,7 +113,7 @@ export const useTotalsChart = (baseCurrency: FiatKeys) => {
     );
   }
 
-  if (currentAPY) {
+  if (currentAPY && !barConfig.find((data) => data.dataKey === 'currentApy')) {
     barConfig.push({
       dataKey: 'currentApy',
       title: <FormattedMessage defaultMessage="Current APY" />,
```
jeffywu: Is the tooltip title wrong here?

packages/features/portfolio/src/containers/portfolio-overview/hooks/use-totals-chart.tsx
```diff
@@ -77,7 +77,10 @@ export const useTotalsChart = (baseCurrency: FiatKeys) => {
     [themeVariant, baseCurrency, netWorth]
   );
 
-  if (debts?.isNegative()) {
+  if (
+    debts?.isNegative() &&
+    !barConfig.find((data) => data.dataKey === 'totalAssets')
```
jeffywu: Do we have to do this check because of the useMemo above? what if we remove that as well?

packages/features/portfolio/src/containers/portfolio-overview/portfolio-overview.tsx
matthew-garrett: I updated the check to when we're setting the totalVaultHoldings. Lmk what you think

packages/features/portfolio/src/containers/portfolio-overview/portfolio-overview.tsx
```diff
@@ -32,6 +32,8 @@ const PortfolioOverview = () => {
     vaultHoldingsData.length === 0 &&
     riskOverviewData.length === 0;
 
+  console.log('vaultHoldingsData: ', vaultHoldingsData);
```
jeffywu: Remove this log

packages/features/portfolio/src/hooks/use-vault-holdings-table.tsx
```diff
@@ -479,7 +481,7 @@ export const useVaultHoldingsTable = () => {
     }
   }, [expandedRows, setExpandedRows]);
 
-  if (totalVaultHoldings) {
+  if (totalVaultHoldings && !totalVaultHoldings.amountPaid.isZero()) {
```
jeffywu: I don't think you can rely on this as a filter if the subgraph is down, you should actually add something that counts the total positions inside the holdings

PR #1265, opened 2024-11-15
Perf/remove notional context

matthew-garrett:
- Removes notional context


packages/features/portfolio/src/portfolio-feature-shell.tsx
```diff
@@ -45,12 +45,10 @@ export const PortfolioFeatureShell = observer(() => {
   const network = useSelectedNetwork();
   const isAccountLoading = useAccountLoading();
   const isAcctAndBalanceReady = useAccountAndBalanceReady(network);
-  const { hasNoteOrSNote } = usePortfolioNOTETable();
```
jeffywu: Are we supposed to remove this?

matthew-garrett: I'll add it back. It doesn't hurt anything that I can see. It just seemed unnecessary and changed nothing when I removed it but maybe there is a edge case I am missing.

PR #1273, opened 2024-11-20
build: adds separate validBoostDate

matthew-garrett:


packages/notionable/src/global/account/communities.ts
```diff
@@ -100,11 +100,17 @@ export function checkSanctionedAddress(account: string) {
 }
 
 export async function checkNewUserAddress(account: string) {
-  if (new Date() > boostEndDate) return false;
-
   const eligibleAddresses = await fetchNewcomerBoostData();
 
   return eligibleAddresses
     ?.map((addr) => addr.toLowerCase())
     .includes(account.toLocaleLowerCase());
```
jeffywu: Use regular lower case

PR #1284, opened 2024-11-27
Perf/trade model access

matthew-garrett:
- Removes old context from Lend, Borrow and Provide liquidity. 
- Adds new trade model data from MobX


packages/shared/trade/src/borrow-terms/use-borrow-terms.tsx
```diff
@@ -35,7 +35,11 @@ export const useBorrowTerms = (
   const nonLeveragedYields = currentNetworkStore.getAllNonLeveragedYields();
 
   const spotMaturityData = useSpotMaturityData(
-    deposit ? availableDebtTokens : []
+    deposit
```
jeffywu, line 38: Does this selector get the right type?
image

matthew-garrett: deposit comes through as  TokenDefinition | undefined. This is a only used on the leveraged pages that are known to be broken at this time.

PR #1289, opened 2024-12-05
build: wip

matthew-garrett:
build: updates mobile nav styling and icons


packages/shared/wallet/src/network-selector/network-selector.tsx
```diff
@@ -138,16 +156,181 @@ export function PortfolioNetworkSelector({
       sx={sx}
       isPortfolio
       onNetworkChange={(network) => portfolioStore.setNetwork(network)}
+      hideOnMobile
     />
   );
 }
 
+export function MobileNetworkSelector() {
```
jeffywu: Should we split this into a different file?

packages/shared/wallet/src/network-selector/network-selector.tsx
```diff
@@ -138,16 +156,181 @@ export function PortfolioNetworkSelector({
...
+            boxShadow: theme.shape.shadowLarge(),
+            border: theme.shape.borderStandard,
+            background: theme.palette.background.default,
+          },
+        }}
+      >
+        <NetworkInnerWrapper
+          hideNetWorth={hideNetWorth}
+          theme={theme}
+          sx={{ marginTop: '72px', padding: '16px !important' }}
```
jeffywu: Can we use theme.spacing throughout here?

packages/shared/web/src/header/header.tsx
```diff
@@ -60,6 +63,19 @@ export function Header({ children }: HeaderProps) {
     setInLocalStorage('hideSubGraphError', true);
   };
 
+  useEffect(() => {
```
jeffywu, line 66: Should this be done in the App.tsx file instead?

matthew-garrett: Imo we keep this in the header. We want the local state update happening in the header component. It guarantees that the mobile nav is applied immediately.

packages/shared/web/src/header/header.tsx
```diff
@@ -60,6 +63,19 @@ export function Header({ children }: HeaderProps) {
...
+      setIsMobileView(false);
+    }
+    const handleResize = () => {
+      setIsMobile(window.innerWidth <= 768);
+    };
+    window.addEventListener('resize', handleResize);
+    return () => window.removeEventListener('resize', handleResize);
+  }, [setIsMobileView, isMobileView]);
+
   window.addEventListener('scroll', () => {
```
jeffywu: Should this also go into a useEffect and have a removeEventListener as well?

PR #1331, opened 2025-01-22
make transaction history responsive in mobile

amirnazari96:
Make the tansaction history tab responsive in mobile
Fix the header responsiveness when switching between desktop and mobile view

### DIFF
--- a/packages/common/mui/src/lib/simple-toggle/simple-toggle.tsx
+++ b/packages/common/mui/src/lib/simple-toggle/simple-toggle.tsx
@@ -59,7 +59,7 @@ export const SimpleToggle = ({
 const Container = styled(Box)(
   ({ theme }) => `
   height: 100%;
-  background: ${theme.palette.background.paper};
+  background: ${theme.palette.background.button};
   border-radius: ${theme.shape.borderRadius()};
 `
 );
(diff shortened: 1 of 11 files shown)


packages/common/mui/src/lib/simple-toggle/simple-toggle.tsx
```diff
@@ -59,7 +59,7 @@ export const SimpleToggle = ({
 const Container = styled(Box)(
   ({ theme }) => `
   height: 100%;
-  background: ${theme.palette.background.paper};
+  background: ${theme.palette.background.button};
```
jeffywu: @anazari96 This toggle is used in a few other places inside table headers so the color changes here make it look wrong:

Your Branch:
image

On Dev:
image

You can see this toggle when you go to the Earn > Lending tab (or any of these "dashboard card view" pages)

jeffywu: Also on dark mode the text should be white:
image

It looks like it is now darker on your version:
image

You can see this on the "Holdings" view in the Portfolio for the test address I sent you

And text is white now on light mode:
image

jeffywu: You can access dark mode via the settings:
image

PR #1369, opened 2025-02-27
Fix/hooks-desktopchart-drawer

amirnazari96:


packages/features/portfolio/src/containers/portfolio-overview/containers/leverage-vaults-overview.tsx
```diff
@@ -69,6 +69,15 @@ const LeverageVaultsOverview = ({ data }: IProps) => {
   );
 };
 
+const Header = styled(H2)(({ theme }) => ({
```
jeffywu: I will ask Isaac to match this to the style library rather than doing something custom.

jeffywu: @anazari96 switch all the headings into H4

packages/features/portfolio/src/components/portfolio-detail/portfolio-detail.tsx
```diff
@@ -115,7 +116,17 @@ const PortfolioDetail: React.FC<PortfolioDetailProps> = ({
       {buttonData && (
         <ActionButtonContainer>
           {buttonData.map((button, index) => (
-            <Button key={index} variant="outlined" color="primary" fullWidth>
+            <Button
+              key={index}
+              variant="outlined"
+              color="primary"
+              onClick={
```
jeffywu: The button from `notional/mui` should have a `to` prop which uses the react-router-dom `Link` component inside instead. If you're not using that component then we can just wrap this in a Link component instead. I'd prefer to remove all click handlers that call `navigate` because the browser is not aware that it is a link.

packages/features/portfolio/src/side-drawers/withdraw.tsx
```diff
@@ -42,31 +43,41 @@ export const Withdraw = observer(() => {
...
+            message={<FormattedMessage {...messages[warning]['message']} />}
+            maxWidth={'100%'}
+          />
+        )}
+      </PortfolioSideDrawer>
+    </Container>
   );
 });
+
+const Container = styled(Box)(
```
jeffywu: Can we put this inside `TransactionSidebar` or `Drawer` instead? I want all the drawers to appear the same way on mobile.

PR #1461, opened 2025-12-02
Exponent liquidator

jeffywu:


packages/contracts/src/abi/ExponentFlashLiquidator.json
jeffywu, line 1: Remove the abis that you're no longer using.

apps/exponent-liquidator/src/abis/addressRegistry.ts
jeffywu, line 1: We can move a lot of these into the actual contracts/src package and then we will get better types.

apps/exponent-liquidator/src/configs/vaultConfig.ts
```diff
@@ -0,0 +1,60 @@
...
+  },
+  [Network.arbitrum]: {
+    // Arbitrum vault configurations
+  },
+} as const;
+
+// Vault liquidation settings by network and vault address
+export const VaultLiquidationSettings = {
+  [Network.mainnet]: {
+    '0x7f723fee1e65a7d26be51a05af0b5efee4a7d5ae': {
```
jeffywu: For scalability you should set default parameters here and then only set overrides.

apps/exponent-liquidator/src/utils/morphoRouter.ts
```diff
@@ -0,0 +1,166 @@
...
+    const healthFactorData: HealthFactorData[] = [];
+
+    for (let i = 0; i < pairs.length; i++) {
+      const [account, vault] = pairs[i];
+      const healthData = results[`health_${i}`] as [
+        BigNumber,
+        BigNumber,
+        BigNumber
+      ];
+      const [borrowed, , maxBorrow] = healthData;
```
jeffywu: The collateral shares is already returned here, you don't need to execute another batch call.

apps/exponent-liquidator/src/utils/morphoRouter.ts
```diff
@@ -0,0 +1,166 @@
...
+    );
+  }
+
+  async batchHealthFactors(pairs: Position[]): Promise<HealthFactorData[]> {
+    const calls: AggregateCall[] = pairs.map(([account, vault], index) => ({
+      stage: 0,
+      target: this.morphoRouterContract,
+      method: 'healthFactor',
+      args: [account, vault],
+      key: `health_${index}`,
```
jeffywu, line 48: You can pass a transform function here on each call instead of doing a loop below.

apps/exponent-liquidator/src/ExponentLiquidator.ts
```diff
@@ -0,0 +1,851 @@
...
+    ) as TradingModule;
+  }
+
+  async getRiskyPositions(): Promise<RiskyPosition[]> {
+    // Step 1: Batch healthFactor calls to get raw health factor data
+    const healthFactorData =
+      await this.morphoRouterIntegration.batchHealthFactors(this.positions);
+
+    // Step 2: Batch borrowShares calls to get borrow share data
+    const borrowSharesData =
```
jeffywu: You don't need this data here, you only need it for the accounts that have risky positions. I would not execute for all accounts here.

apps/exponent-liquidator/src/ExponentLiquidator.ts
```diff
@@ -0,0 +1,851 @@
...
+          withdrawRequestStatuses[index].secondaryWithdrawTokenAmount,
+        accountVaultSharePrice,
+      };
+    });
+  }
+
+  private calculateLiquidationAmounts(
+    positions: EnrichedPosition[],
+    liquidationIncentiveFactor: ethers.BigNumber
+  ): {
```
jeffywu: Return type declarations like this are not necessary because of type inference.

apps/exponent-liquidator/src/ExponentLiquidator.ts
```diff
@@ -0,0 +1,851 @@
...
+  }
+
+  private sortPositionsForLiquidation(positions: EnrichedPosition[]): Map<
+    string,
+    {
+      withoutWithdrawRequest: EnrichedPosition[];
+      withWithdrawRequest: EnrichedPosition[];
+    }
+  > {
+    const sortedByVault = new Map<
```
jeffywu: This can be rewritten as a reduce operation that will read a lot more clear, in my opinion. There are a lot of re-declared type declarations that muddy up what is actually happening here.

package.json
```diff
@@ -28,6 +28,11 @@
     "@safe-global/safe-apps-provider": "^0.18.3",
     "@safe-global/safe-apps-sdk": "^9.1.0",
     "@segment/analytics-next": "^1.58.0",
+    "@stablelib/chacha20poly1305": "^2.0.1",
```
jeffywu: I'm not sure these dependencies are necessary, you should be able to sign transactions and do all the cryptographic stuff via ethers.

apps/exponent-liquidator/src/abis/addressRegistry.ts
T-Woodward: Why do you want me to do this? So that they become generally available to other apps in the repo?

apps/exponent-liquidator/src/utils/morphoRouter.ts
T-Woodward: healthFactor returns collateralValue, not collateralShares. Would have to get the vaultshare price by borrower and use it to transform the result to get it back to what I want

package.json
T-Woodward: Ok I'm not sure what to do in response to this comment. I think those are in there because I had to install them specifically to get it to work.

